PR #1, opened 2018-02-13
EZP-28009: Created RichText FieldType Bundle

alongosz:
This PR introduces basic skeleton for RichText FieldType Bundle.
This is only the first step towards [EZP-28009](https://jira.ez.no/browse/EZP-28009), it allows adding the Bundle to `AppKernel` (with DI config placeholder).

The directory structure is as follows:
```
.
+-- src
|   +-- bundle
|   +-- lib # not available yet, future PR will introduce it
+-- tests # also not available yet
|   +-- lib
|   +-- bundle
```

**Introduced:**
- [x] Bundle class.
- [x] Extension class with injected placeholder for services configuration (`services.yml`).
- [x] README, LICENSE.
- [x] Composer (needed also by Packagist for README installation instructions to work).
- [x] php-cs-fixer configuration
- [x] Initial PHPUnit configuration.

Please note that links included in README will not work yet as adding all Resources from `ezpublish-kernel` RichText FieldType would clutter this PR diff.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,37 @@
+<?php
+
+$header = <<<'EOF'
+@copyright Copyright (C) eZ Systems AS. All rights reserved.
+@license For full copyright and license information view LICENSE file distributed with this source code.
+EOF;
+
+return PhpCsFixer\Config::create()
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'concat_space' => ['spacing' => 'one'],
+        'array_syntax' => ['syntax' => 'short'],
+        'simplified_null_return' => false,
+        'phpdoc_align' => false,
+        'phpdoc_to_comment' => false,
+        'cast_spaces' => false,
+        'blank_line_after_opening_tag' => false,
+        'single_blank_line_before_namespace' => true,
+        'space_after_semicolon' => false,
+        'header_comment' => [
+            'commentType' => 'PHPDoc',
+            'header' => $header,
+            'location' => 'after_open',
+            'separate' => 'top',
+        ],
+        'yoda_style' => false,
+        'no_break_comment' => false,
+        'declare_strict_types' => true,
+    ])
+    ->setRiskyAllowed(true)
+    ->setFinder(
+        PhpCsFixer\Finder::create()
+            ->in(__DIR__ . '/src')
+            ->files()->name('*.php')
+    )
+;
--- a/.php_cs
+++ b/.php_cs
@@ -31,7 +31,7 @@ return PhpCsFixer\Config::create()
     ->setRiskyAllowed(true)
     ->setFinder(
         PhpCsFixer\Finder::create()
-            ->in(__DIR__ . '/src')
+            ->in([__DIR__ . '/src', __DIR__ . '/tests'])
             ->files()->name('*.php')
     )
 ;
--- a/README.md
+++ b/README.md
@@ -1,2 +1,57 @@
-# ezplatform-richtext-fieldtype
-eZ Platform Rich Text Field Type Bundle
+# eZ Platform RichText Field Type Bundle
+
+This Bundle provides RichText (`ezrichtext`) Field Type for eZ Platform 2.1 and higher.
+It is a Field Type for supporting rich formatted text stored in a structured XML format.
+
+This package overrides the `eZ\Publish\Core\FieldType\RichText` namespace provided by
+the `ezsystems/ezpublish-kernel` package.
+
+This Field Type succeeds the former [XMLText](https://github.com/ezsystems/ezplatform-xmltext-fieldtype)
+Field Type found in eZ Publish 5.x and before.
+
+## Background
+
+When looking to find a structured text format for eZ Platform, we wanted to pick something that
+was widely used in the industry and which could support the custom & embed structures we have
+had in eZ Publish for years, which has enabled us to seamlessly target several channels / formats
+based on the same internal stored formats.
+
+What we had at the time was inspired by early drafts of XHTML 2.0, a standard made for the most
+part obsolete by html5.
+
+We also knew from experience we had to support html5 as an input/output format for RichText editors
+to reduce the number of customizations we had to apply on top of available editors. Which would make
+it hard to keep up to date, and forces us to deal with edge cases ourselves instead of relying on
+the editor doing it for us.
+
+In RichText we have ended up with a solution that is built on a more widely used internal format,
+moved closer to html5 supported by editors, and better suited to support wider range of formats.
+
+## Format
+
+### Storage format
+
+Storage format in RichText is [DocBook](http://docbook.org/), for further info on its schema and how we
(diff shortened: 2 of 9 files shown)


.php_cs
```diff
@@ -0,0 +1,37 @@
...
+        'space_after_semicolon' => false,
+        'header_comment' => [
+            'commentType' => 'PHPDoc',
+            'header' => $header,
+            'location' => 'after_open',
+            'separate' => 'top',
+        ],
+        'yoda_style' => false,
+        'no_break_comment' => false,
+        'declare_strict_types' => true,
```
alongosz: Note: this is new in our bundles, along with `single_blank_line_before_namespace => true` it enforces
```php
declare(strict_types=1);

namespace EzSystems\EzPlatformRichTextFieldTypeBundle\DependencyInjection;
```
after the licensing comment in each PHP file.

README.md
```diff
@@ -1,2 +1,84 @@
-# ezplatform-richtext-fieldtype
```
alongosz: @DominikaK README review-please-ping ;)

composer.json
```diff
@@ -0,0 +1,35 @@
...
+        "psr-4": {
+            "EzSystems\\Tests\\EzPlatformRichTextFieldType\\": "tests/lib"
+        }
+    },
+    "scripts": {
+        "fix-cs": "@php ./vendor/bin/php-cs-fixer fix -v --show-progress=estimating"
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "0.1.x-dev"
```
lserwatka: Please start from 1.0.x-dev, Rich Text is mature.

alongosz: @lserwatka fixed :+1:

lserwatka:
We need CI setup for this repo. Ping help @mnocon

README.md
```diff
@@ -1,2 +1,84 @@
...
+
+3. Remember to clear the Symfony Cache (for `SYMFONY_ENV` your project uses).
+    ```bash
+        php bin/console cache:clear
+    ```
+
+## Background
+
+When looking to find a structured text format for eZ Platform, we wanted to pick something that
+was wildly used in the industry and which could support the custom & embed structures we have
```
DominikaK: `widely used`, not `wildly` ;)

andrerom:
Will we be able to import git history (git subtree split magic as Bertrand and others have done before) of the field type from kernel still even if we make the skeleton first? _(Context: We frequently need this for support to go back and see why or when something was changed)_

README.md
alongosz: @DominikaK thanks, amended in d271944

alongosz:
> Will we be able to import git history (git subtree split magic as Bertrand and others have done before) of the field type from kernel still even if we make the skeleton first? (Context: We frequently need the history for support to go back and see why or when something was changed)

First of all, I haven't done that before, so I might be wrong, but I see several problems with that:
1. The structure of the Bundle is different, please notice that we don't have EzRichTextFieldTypeBundle in ezpublish-kernel. I moved files around.
2. AFAICS subtree is meant to split active and maintained part of a project as a subdirectory. Because of [1.] this is not really a subdirectory.
3. `Core/FieldType/RichText` have grown over the years, I intend to do a cleanup (more moving files around, e.g. utilizing Sf 3.4+ DI Container configuration features).
4. Namespaces are different, so that would need to be changed in moved files anyway.
5. Related to [2.] and [3.] the Bundle code is intended to diverge from what we have in `Core/FieldType/RichText` namespace. The `Core` namespace is going to be deprecated. AFAICS Subtree is something that can be pulled-in when upstream changes. This is not the case here.
6. Doing a split would mean having files initially put in a wrong place (the code wouldn't work) and then applying proper changes as a separate commits. IMHO it would rather clutter history.

**Summary**: I'm not a huge fan of the idea. I understand advantages provided by continuous history, however IMHO that would lead to cluttering it and applying more work trying to cleanup instead of introducing a correct code in the first place. 
Moreover, if I read the [Git subtree doc](https://github.com/git/git/blob/v2.7.4/contrib/subtree/git-subtree.txt#L22-L64) right, while it is possible to diverge a subtree that has been split, it's rather intented for merges, not for code that somewhat introduces new stack / structure.

POV ping @lserwatka @bdunogier @Nattfarinn

andrerom:
1. this is possible with some additional magic.
2, magic, lots of things is possible, including extracting several subtrees and importing them in a common code base
3. can be done after the import
4. same, but can maybe also be rewritten during the split afaik, which would be cleaner as it will then adapt the history afaik
5. In a split scenario we remove `Core/FieldType/RichText` and have BC note that the field type has been moved out to own package with adjusted namespace.
6.some history is better then none, initial comkit blobs are the worst

Don't dismiss the magic you don't believe to exist or don't have knowledge of, magic is real 🎆  😄

Nattfarinn:
> Any sufficiently advanced technology is indistinguishable from magic.
> -- Arthur C. Clarke

bdunogier:
To be honest, unless we figure out how to keep history without spending days on it, I'm not gonna _insist_ on it.

The history does exist, inside eZ Publish kernel. Maybe the commits where the code is moved from the kernel could have a link to the reference this was splitted from ? And the branch it still lives in. That way, we can still go through history, even if there's an extra jump.

If we _really_ want to make it happen, then we're gonna have to get creative :)
Maybe it would be simpler to do the opposite: prune from ezpublish-kernel everything that it NOT richtext, and keep the history for what's left.

Then, merge that into this repository (or use it as the initial commit of it).

alongosz:
> Maybe it would be simpler to do the opposite: prune from ezpublish-kernel everything that it NOT richtext, and keep the history for what's left.

Filtering branch is painful :) and in case of RichText not as simple as it would seem.
While working with that I discovered that parts of RichText are scattered across ezpublish-kernel (`ez\Publish\Core\FieldType\RichText` namespace is not everything that makes RichText work).

Also that would mean overriding what already got merged.

However, I managed to split subtree of the mentioned namespace. It could be added as a bunch of separate commits. But because there are parts outside of this namespace, making more splits would be even more time consuming. In some cases (e.g. yaml configuration, Twig helper) it's not possible to make that split at all w/o keeping unrelated code).

This FieldType needs a cleanup (adjustments to having code in a single bundle instead of CoreBundle + Kernel Core - some parts need to go away because of a bundle architecture). So for me it's rather a new code than a modified one.

> Maybe the commits where the code is moved from the kernel could have a link to the reference this was splitted from

To summarize: I'm more in favor of this ^ solution.

I'll show you in coming days PR with extracted subtree so you can decide if you prefer it squashed as I do.

bdunogier:
The most important is I think that we have traceable history for it. If the commit that adds the files for the first time has links to the point in ezpublish-kernel's history where it was extracted from, I think we are okay.

One note about the various sources for the files in the kernel: that's exactly why I suggested that you may want to start with ezpublish-kernel, and remove everything that isn't richtext. You end up with a structure that matches what was in the kernel before, but you can use `git move` to rename the files. Then we keep a full history. Maybe for another time :)

PR #5, opened 2018-04-13
EZP-28009: Changed priority of kernel pass to avoid dealing with tags

bdunogier:
By deleting kernel services before their tags are processed, we don't need to remove them from the services they're added to as parameters.

| Question           | Answer
| **JIRA issue**     | None
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Increases the priority of the kernel pass. Ensures that services we remove haven't been processed for tags, and haven't been injected into other services as arguments.

Fixes the problems reported by QA on ezsystems/ezpublish-kernel#2302.

### DIFF
--- a/src/bundle/DependencyInjection/Compiler/KernelRichTextPass.php
+++ b/src/bundle/DependencyInjection/Compiler/KernelRichTextPass.php
@@ -10,7 +10,6 @@ namespace EzSystems\EzPlatformRichTextBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Handles compatibility with kernel 7.x.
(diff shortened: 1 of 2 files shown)


src/bundle/DependencyInjection/Compiler/KernelRichTextPass.php
```diff
@@ -30,31 +29,13 @@ public function process(ContainerBuilder $container)
         array_map(
             function ($service) use ($container) {
                 if ($container->hasDefinition($service)) {
+
                     $container->log($this, "Removed ezpublish-kernel richtext service: $service");
                     $container->removeDefinition($service);
+                    echo "Removing $service\n";
```
adamwojs: ?

bdunogier: Oops :)

alongosz:
Merged as-is. At least one more step is required before we can pass this package to QA.

PR #7, opened 2018-10-12
[WiP] Applied changes from Kernel since extracting RichText

alongosz:
| Question           | Answer
| **JIRA issue**     | Part of [EZP-28009](https://jira.ez.no/browse/EZP-28009)

This PR shows changes to RichText FieldType that happened since it was extracted at ezsystems/ezpublish-kernel@ab3855d7d1f99959849bd2da81452d60a36144f6.

Changes include:
- [x] ezsystems/ezpublish-kernel#2386
- [x] ezsystems/ezpublish-kernel#2381
- [x] ezsystems/ezpublish-kernel#2351
- [ ] Custom Styles

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -26,8 +26,4 @@ if (!($settings = include(__DIR__ . '/config.php'))) {
     throw new \RuntimeException('Could not read config.php, please copy config.php-DEVELOPMENT to config.php & customize to your needs!');
 }
 
-// Class alias used for BC
-// Enables old code which still extends non namespaced TestCase to work
-class_alias('PHPUnit\Framework\TestCase', 'PHPUnit_Framework_TestCase');
-
 require_once __DIR__ . '/vendor/autoload.php';
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -11,9 +11,7 @@
   <testsuites>
     <testsuite name="ezrichtext_field_type">
       <directory>tests/lib</directory>
-      <exclude>
-        <directory>tests/lib/eZ/API</directory>
-      </exclude>
+      <exclude>tests/lib/eZ/API</exclude>
     </testsuite>
   </testsuites>
 </phpunit>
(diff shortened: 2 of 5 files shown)


andrerom:
4 reasons and counting for leaving this bundle for 8.x ? 😃

alongosz:
> 4 reasons and counting for leaving this bundle for 8.x and up ?

Nooo :) I'm just running tests here. Failure is not an option, too many Stories depend on that.
What we're going to do in 8.x is to remove RichText from Kernel :stuck_out_tongue:

andrerom:
Point was having it in kernel and in bundle at the same time is messy, so I hope we won't use this in the product before 3.0 when removed in kernel. But sounds like that is your plan more or less so 👍

alongosz:
> Point was having it in kernel and in bundle at the same time is messy, so I hope we won't use this in the product before 3.0 when removed in kernel.

For some of the features (like a few most common built-in Custom Tags) we'd like to use eZ Design. It cannot be done in Kernel because of cyclic dependency of eZ Design on Kernel. So the current plan is to have it in 2.4. // cc @SylvainGuittard

andrerom:
ok, that makes sense. From support perspective I'm just concerned that having it in kernel and bundle at the same time will be messy for us and contributors. But as 8.0 kernel comes not so much later, and it will be basically removed in master a few months after, I guess we'll manage.

lserwatka:
Yes, it is manageable. We are fully aware of the consequences but are enforced by semver and BC.

bdunogier:
About having it both in kernel and outside, the only way not to is to keep it in the kernel. The same argument will be valid for anything we'd intend to extract. As far as I know, this is nothing new.

Note though that given how the splitting was done, applying patches across repos should be much easier than before. The package's [initial commit](https://github.com/ezsystems/ezplatform-richtext/commit/37121d2466faa47e3e0d7a7cf85801b4b8c44a94) was done by copying the files from kernel with the exact same path. They were then renamed in [further commits](https://github.com/ezsystems/ezplatform-richtext/commit/cd920c92ac4ec2561526f6ad65b537f15935984c).

@alongosz himself was surprised to see that fixes on kernel applied to ezplatform-richtext-fieldtype without requiring any manual change.

alongosz:
Indeed. So as long as fix applies to already existing files, cherry-pick is very smooth. Custom Tags required more work because of new files added. But as it goes for maintenance-only of RichText in Kernel, I actually expect simple fixes.

PR #10, opened 2018-10-29
[Travis] Enabled Solr integration tests

alongosz:
| Question           | Answer
| **JIRA issue**     | Related to [EZP-28009](https://jira.ez.no/browse/EZP-28009)
| **Bug/Improvement**| yes, improvement
| **New feature**    | no
| **Target version** | `1.0`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

This PR refactors existing SetupFactories and enables integration testing using Solr.

Since there is no way to directly reuse existing SetupFactory in Solr tests (because we'd miss required RichText package overrides), it's a good chance to refactor approach used by kernel SetupFactory which relies on `config.php` and `containerBuilder.php`.

`SetupFactory` implemented for RichText no longer requires mentioned files, taking entirely responsibility of building Service Container for tests.

In the end some code needs to be moved to ezpublish-kernel to avoid hardcoding its path, but since other packages also rely on kernel SetupFactory, it's too much to be handled right now.

**TODO**:
- [x] Refactor `SetupFactory` to stop relying on `config.php`
- [x] Implement Solr Setup Factory which takes into the account RichText package instead of Core RichText implementation.
- [x] Enable Solr integration tests on Travis.
- [x] Wait for Travis to confirm
- [x] Ask for Code Review.


alongosz:
Since there are no objections - merging :stuck_out_tongue_winking_eye:

PR #23, opened 2018-12-06
EZP-29812: As an editor I want to embed a content item within a block element

alongosz:
| Question           | Answer
| **JIRA issue**     | [EZP-29812](https://jira.ez.no/browse/EZP-29812)
| **Required by** | ezsystems/ezplatform-admin-ui#736
| **Bug/Improvement**| no
| **New feature**    | yes
| **Target version** | `1.0 (master) for eZ Platform 2.4`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | yes

This PR adds support for `embed-inline` View Type.

**TODO**:
- [x] Refactor prepending core DIC settings for better readability.
- [x] Create and configure `embed-inline` eZ Platform View Type (SA-aware).
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
+++ b/src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
@@ -9,11 +9,13 @@ declare(strict_types=1);
 namespace EzSystems\EzPlatformRichTextBundle\DependencyInjection;
 
 use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\Config\Resource\FileResource;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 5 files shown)


src/bundle/Resources/views/themes/standard/content_view/embed-inline.html.twig
```diff
@@ -0,0 +1,7 @@
+{% set content_name=ez_content_name(content) %}
```
lserwatka: Can't we use just `{{ content.name }}`?

alongosz: Great catch, SA-aware repo is indeed injected when rendering it. Fixed via fixup 5b28f0f. I also changed template name to `embed_inline.html.twig` so it follows existing naming convention for templates.

PR #32, opened 2019-04-01
EZP-30166 :  RichText Link converter emits warning during custom tag conversion

vidarl:
| Question           | Answer
| **JIRA issue**     | [EZP-30166](https://jira.ez.no/browse/EZP-30166)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 2.4?
| **BC breaks**      | no
| **Tests pass**     | no
| **Doc needed**     | no

The cause of the problem seems to be two things:
* In https://github.com/ezsystems/ezplatform-richtext/commit/23bd49b3797319f3ae2350ba1536dfd9a1e6dd31#diff-7310606d4fef503a40dee782da35ae00R210 we create a DOM document with multiple root notes
* DOMXPath seems to only take namespace definitions from the first parent element, and use the same ones on any other root elements. 

So in https://github.com/ezsystems/ezplatform-richtext/blob/master/src/lib/eZ/RichText/Converter/Link.php#L67, if the DOMDocument looks like this:

```
<?xml version="1.0"?>
<para xmlns="http://docbook.org/ns/docbook">Just a regular text</para>
<para xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"><link xlink:href="" xlink:show="none">ezlocation URL</link></para>
<para xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"><link xlink:href="ezurl://1" xlink:show="none" xlink:title="">ezurl URL </link></para>
```

Then the namespace in `<link xlink:href=""(..)` will not be recognized and we'll get the reported behaviour : `An exception has been thrown during the rendering of a template ("Warning: DOMXPath::query(): Undefined namespace prefix") in "@admin/fieldtypes/preview/content_fields.html.twig".`

If you would swap the two firsts two `<para>`, then the warning will not be thrown, for the reasons explained above:

```
<?xml version="1.0"?>
<para xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"><link xlink:href="" xlink:show="none">ezlocation URL</link></para>
<para xmlns="http://docbook.org/ns/docbook">Just a regular text</para>
<para xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"><link xlink:href="ezurl://1" xlink:show="none" xlink:title="">ezurl URL </link></para>
```

My fix is to render each root node one-by-one. AFAIK, that should not change the end result

**TODO**:
- [x] Fix tests tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Link.php
+++ b/src/lib/eZ/RichText/Converter/Link.php
@@ -58,8 +58,11 @@ class Link implements Converter
     public function convert(DOMDocument $document)
     {
         $document = clone $document;
+
         $xpath = new DOMXPath($document);
         $xpath->registerNamespace('docbook', 'http://docbook.org/ns/docbook');
+        $xpath->registerNamespace('xlink', 'http://www.w3.org/1999/xlink');
+
         $linkAttributeExpression = "starts-with( @xlink:href, 'ezlocation://' ) or starts-with( @xlink:href, 'ezcontent://' )";
         $xpathExpression = "//docbook:link[{$linkAttributeExpression}]|//docbook:ezlink";
 
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -186,6 +186,9 @@ class Template extends Render implements Converter
     {
         $innerDoc = new DOMDocument();
 
+        $rootNode = $innerDoc->createElementNS('http://docbook.org/ns/docbook', 'section');
+        $innerDoc->appendChild($rootNode);
+
         /** @var \DOMNode $child */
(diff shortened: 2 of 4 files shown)


vidarl:
This fix causes problems with some of the Mocks, so a few tests will fail.. Needs to be fixed

micszo:
Shouldn't base be changed to `1.0` since customer versions are `2.4`?

lserwatka:
Could you merge it up?

adamwojs:
Done.

PR #34, opened 2019-04-08
EZP-29876: New break lines disappear in Richtext while styled with custom styles

adamwojs:
| Question           | Answer
| **JIRA issue**     | [EZP-29876](https://jira.ez.no/browse/EZP-29876)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `1.0`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Note for QA: This bugfix depends on https://github.com/ezsystems/ezplatform-richtext/pull/32

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
@@ -657,7 +657,7 @@
     <xsl:apply-templates select="node()|@*"/>
   </xsl:template>
 
-  <xsl:template match="docbook:eztemplate[@type='style']/docbook:ezcontent/text()">
+  <xsl:template match="docbook:eztemplate/docbook:ezcontent/text()">
     <xsl:call-template name="breakLine">
       <xsl:with-param name="text" select="."/>
     </xsl:call-template>
--- a/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
@@ -657,6 +657,12 @@
     <xsl:apply-templates select="node()|@*"/>
   </xsl:template>
 
+  <xsl:template match="docbook:eztemplate[@type='style']/docbook:ezcontent/text()">
+    <xsl:call-template name="breakLine">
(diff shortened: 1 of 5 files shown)


tests/lib/eZ/RichText/Converter/Xslt/_fixtures/docbook/031-ezstyle.xml
```diff
@@ -13,4 +13,11 @@
     <eztemplate name="highlighted_block" type="style" ezxhtml:align="center">
         <ezcontent>Highlighted block with "center" align.</ezcontent>
     </eztemplate>
+    <eztemplate name="highlighted_block" type="style">
+        <ezcontent>
+            Highlighted block with multiple lines
+            Highlighted block with multiple lines
```
alongosz: This doesn't seem like a correct XML. I'm not sure if `<br>` is proper for DocBook, but XML spec is rather not white-space-aware.

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
```diff
@@ -657,6 +657,12 @@
     <xsl:apply-templates select="node()|@*"/>
   </xsl:template>
 
+  <xsl:template match="docbook:eztemplate[@type='style']/docbook:ezcontent/text()">
```
alongosz: `@type` is deprecated, we shouldn't rely on this. If the problem exists it probably applies both to Styles and Tags.

adamwojs: >   If the problem exists it probably applies both to Styles and Tags.

Actually, the problem applies only to Custom Styles.  The content of the Custom Tags is one or more block elements (paragraphs, ordered lists, etc) and the line breaks are transformed during processing of those block elements. Nevertheless, fixed in https://github.com/ezsystems/ezplatform-richtext/pull/34/commits/f7b6b1c50eb9afbc9c6d05c83d9637786ce40e9c  😉

lserwatka:
@alongosz could you make a final review, we should proceed with QA for 2.5.1

lserwatka:
Could you merge it up?

adamwojs:
Done.

PR #36, opened 2019-04-18
EZP-30483: Custom tags: provide a way to implement custom logic

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-30483](https://jira.ez.no/browse/EZP-30483)
| **Bug/Improvement**| yes
| **New feature**    | yes
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | yes

With this RP you can easily define a new service, which will be used for custom tag custom logic.

Example usage:
### Extension service:

`src/YourBundle/FieldType/RichText/Converter/Render/Template/Extension/Accordion`: 
```php
namespace YourBundle\FieldType\RichText\Converter\Render\Template\Extension;

use EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Extension\Base;

class Accordion extends Base
{
    // you can inject any services there
    public function __construct(string $identifier, string $type)
    {
        parent::__construct($identifier, $type);
    }

    // injected services can be used here
    public function extend(?array $params): array
    {
        return ['new_params' => 'test'];
    }
}
```
`services.yml`:
```yml
ezrichtext.converter.template.extensions.accordion:
    class: YourBundle\FieldType\RichText\Converter\Render\Template\Extension\Accordion
    tags: [ezpublish.ezrichtext.converter.output.template.extension]
    arguments:
        - 'accordion'
        - 'tag'
```

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/config/fieldtype_services.yml
+++ b/src/bundle/Resources/config/fieldtype_services.yml
@@ -50,11 +50,20 @@ services:
         tags:
             - {name: ezpublish.ezrichtext.converter.output.xhtml5, priority: 0}
 
+    _instanceof:
+        EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Extension\Base:
+            tags: [ezpublish.ezrichtext.converter.output.template.extension]
+
+    ezrichtext.converter.template.extensions:
+        class: EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Collection\Extension
+        arguments: [!tagged ezpublish.ezrichtext.converter.output.template.extension]
+
     ezrichtext.converter.template:
         class: EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template
         arguments:
             - '@ezrichtext.renderer'
             - '@ezrichtext.converter.output.xhtml5'
+            - '@ezrichtext.converter.template.extensions'
             - '@?logger'
         tags:
             - {name: ezpublish.ezrichtext.converter.output.xhtml5, priority: 10}
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -14,6 +14,7 @@ use DOMNode;
 use DOMXPath;
 use EzSystems\EzPlatformRichText\eZ\RichText\Converter;
 use EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render;
+use EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Collection\Extension as ExtensionsCollection;
 use EzSystems\EzPlatformRichText\eZ\RichText\RendererInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
@@ -34,18 +35,26 @@ class Template extends Render implements Converter
     private $logger;
 
     /**
+     * @var \EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Collection\Extension
+     */
+    private $extensionsCollection;
+
+    /**
      * RichText Template converter constructor.
      *
      * @param \EzSystems\EzPlatformRichText\eZ\RichText\RendererInterface $renderer
      * @param \EzSystems\EzPlatformRichText\eZ\RichText\Converter $richTextConverter
+     * @param \EzSystems\EzPlatformRichText\eZ\RichText\Converter\Render\Template\Collection\Extension $extensionsCollection
      * @param \Psr\Log\LoggerInterface $logger
      */
     public function __construct(
         RendererInterface $renderer,
         Converter $richTextConverter,
+        ExtensionsCollection $extensionsCollection,
         LoggerInterface $logger = null
     ) {
         $this->richTextConverter = $richTextConverter;
+        $this->extensionsCollection = $extensionsCollection;
         $this->logger = $logger ?? new NullLogger();
 
         parent::__construct($renderer);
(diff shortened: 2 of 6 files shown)


SylvainGuittard:
@dew326 in case you missed it, we would like to have your feedback here :)

dew326:
From my point of view, it's ok but I'm not PHP dev so I added proper reviewers.

SerheyDolgushev:
In a few words, this PR allows to:
1. Set custom service (extension) for each custom tag/custom style
2. You can inject into extensions any services
3. Each extension should have `extend` method. And that method is responsible for implementing custom logic (interacting with injected services, etc) and return additional parameters (like fetched objects, etc) to the custom tag template.

In case if the controller approach will be used, all of the steps need to be done anyway. And additionally, you need to implement a controller. Which will not be used anywhere else. So the solution this PR provides seems simpler and it simplifies solving the original problem.

SerheyDolgushev:
Meantime, published https://packagist.org/packages/contextualcode/ezplatform-richtext-template-extension

emodric:
I'm always against controllers for such purposes as they add measurable overhead (brought by Symfony's subrequests), so I'd go for the solution in this PR.

I never liked controllers and subrequests for rendering eZ views too, as they tend to be misused too much.

emodric:
As for this implementation, I don't like the fact that constructor arguments are predefined and required by the base abstract class. It adds unnecessary complexity to classes extending it. Those values could e.g. be provided by the concrete class itself via static methods.

SerheyDolgushev:
@emodric you are free to define `identifier` and `type` parameters in extension class if you want to do it:
```php
<?php

namespace AcmeBundle\CustomTagExtension;

use Exception;
use eZ\Publish\API\Repository\LocationService;
use ContextualCode\EzPlatformRichTextTemplateExtension\eZ\RichText\Converter\Render\Template\Extension\Base;

class Location extends Base
{
    /**
     * @param string
     */
    protected $identifier = 'location';
    /**
     * @param string
     */
    protected $type = 'tag';
    /**
     * @var \eZ\Publish\API\Repository\LocationService
     */
    protected $locationService;

    /**
     * @param \eZ\Publish\API\Repository\LocationService
     */
    public function __construct(LocationService $locationService)
    {
        $this->locationService = $locationService;
    }

    /**
     * Returns additional parameters which will be available in the view.
     * You can use injected services here to get them.
     *
     * @param array $params Current set of parameters
     *
     * @return array
     */
    public function extend(?array $params): array
    {
        return ['place_location' => $this->locationService->loadLocation($params['place_location_id'])];
    }
}
```

And in this case you don't need to pass `identifier` and `type` params in extension service definition:
```yaml
services:
    ezrichtext.converter.template.extensions.accordion:
        class: AcmeBundle\CustomTagExtension\Location
        tags: [ezpublish.ezrichtext.converter.output.template.extension]
        arguments:
            - '@ezpublish.api.service.location'
```

But the downside for this is that you won't be able to use this custom tag extension for multiple custom tags/styles.

alongosz:
Sorry for the long delay, I was off for two weeks.

Let me say something that I didn't say clearly: this is not a clean way to provide extension point. There are a lot of issues with this solution, starting with naming, through architecture itself and ending with DX.

We've discussed internally proposed changes with @adamwojs but decided not to point out specific things, not to give false hope that this will be accepted. We just see a necessity to resolve it differently.

Answering question from Slack: "Why controller"?
1. Well, first of all because of DX. Adding single line for each Custom Tag or style:
```yaml
# ...
        controller: 'AcmeBundle\CustomTagController:renderAction'
```
is much simpler than 
```yaml
services:
    ezrichtext.converter.template.extensions.accordion:
        class: AcmeBundle\CustomTagExtension
        tags: [ezpublish.ezrichtext.converter.output.template.extension]
```

Even if we provide Symfony auto-configuration and achieve the shortest form possible:
```yaml
services:
    AcmeBundle\CustomTagExtension: ~
```
this still will be some service "hanging" there without proper context.

2. Controller is the right design pattern here as the purpose of it is to provide "view" and add some custom logic. Controller goes beyond just providing some extra parameters, which adds more value.

3. Controller could be a default way to render Custom Templates (Tags, Styles) allowing to cleanup current implementation (it's a little bit messy mostly due to adding new layers on top of existing ones to keep BC).

There's no way I'm accepting this solution in the current form. I hope we will find time to handle it properly. 
The only scenario where we'd refine this code and accept it is if we would find technical difficulties with implementing custom controller support (I get Edi's point about overhead). 

On the other hand we might come up with lightweight controller - just refactoring code and giving responsibility-related names, w/o entire subrequest overhead. We'll see what we can do.

SerheyDolgushev:
So seems like this PR need to be closed. Meantime, we will continue to maintain https://ezplatform.com/packages/contextualcode-ezplatform-richtext-template-extension, untill a similar solution will be released.
Is it possible to make the "new custom tag controller" compatible with https://gitlab.com/contextualcode/ezplatform-richtext-template-extension/blob/master/src/lib/eZ/RichText/Converter/Render/Template/Extension/Base.php#L59-62?

emodric:
Since we're providing opinions, @alongosz, I don't agree with your assesment of this PR. It is ofcourse your right not to merge this PR in this form, but:

> Well, first of all because of DX.

DX does not suffer here with auto-wiring and auto-configuration, since by default, all services are private and, even better, inlined if possible, so there is no low hanging services with no context in the compiled container. Further, if you use `registerForAutoConfiguration()` in DI extension, custom tag extensions could be registered without a single service definition line of code.

> Controller is the right design pattern here as the purpose of it is to provide "view" and add some custom logic. Controller goes beyond just providing some extra parameters, which adds more value.

As I already said, in our experience from various small and large projects, abusing controllers for rendering every single piece of information adds *significant* overhead, so in our book that makes the controller unfit design pattern for this purpose, even if maybe technically correct. The only thing that makes controllers worthwhile here is if custom tags would support ESI caching. And even that too can be achieved without whole subrequest stack running.

SerheyDolgushev:
Sorry, but seems like this PR can be closed. It has been moved to https://ezplatform.com/packages/contextualcode-ezplatform-richtext-template-extension bundle.

alongosz:
> @alongosz sorry, but seems like this PR can be closed. It has been moved to https://ezplatform.com/packages/contextualcode-ezplatform-richtext-template-extension bundle.

Ok. We'll propose our own solution for [EZP-30483](https://jira.ez.no/browse/EZP-30483) once we'll have free resources for that.

PR #39, opened 2019-05-17
EZP-29301: Provided support for custom classes and data attributes

alongosz:
| Question           | Answer
| **JIRA issue**     | [EZP-29301](https://jira.ez.no/browse/EZP-29301)
| **Required by** | ezsystems/ezplatform-admin-ui#996
| **Requires** | ezsystems/ezpublish-kernel#2665
| **Bug/Improvement**| no
| **New feature**    | yes
| **Target version** | `1.1.2` for eZ Platform `2.5.2`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | yes

This PR provides SiteAccess-aware semantic configuration for custom classes and attributes of RichText elements. AlloyEditor integration is provided via ezsystems/ezplatform-admin-ui#996

## Example
A sample configuration is as follows:
```yaml
ezpublish:
    system:
        <scope>: # e.g. default or admin_group
            fieldtypes:
                ezrichtext:
                    classes:
                        paragraph:
                            choices: ['class1', 'class2']
                            required: true
                            default_value: 'class1'
                            multiple: true
                        heading:
                            choices: ['class3', 'class4']
                            required: false
                            default_value: ''
                            multiple: false
                    attributes:
                        paragraph:
                            'select-single-attr':
                                type: 'choice'
                                multiple: false
                                required: true
                                choices: ['value1', 'value2']
                                default_value: 'value2'
                            'boolean-attr':
                                type: 'boolean'
                                default_value: true
                        heading:
                            'text-attr':
                                type: 'string'
                                default_value: 'foo'
                                required: false
                            'number-attr':
                                type: 'number'
                                default_value: 1
                                required: true
```
This configuration is used by ezplatform-admin-ui to add a possibility to pick a class for an element or or set a data attribute to an element. As a result, output HTML would have `class` and `data-<attribute_name>` properties respectively. For more details please see ezsystems/ezplatform-admin-ui#996.

_Please note that setting this configuration for a site (non-admin) scope has no effect_

The list of elements includes:
- embedinline
- embed
- formatted
- heading
- embedimage
- ul
- ol
- li
- paragraph
- table
- tr
- td

## Validation
1. Each custom CSS class config must define `choices` list.
2. A `default_value` of custom class must be the one from `choices` list.
3. Custom data attributes settings must define for each attribute a `type` as one of the `choice`,  `boolean`, `string`, `number`.
4. If a custom data attribute is not of `choice` type, it must not define `choices` list.
5. A `default_value` of custom data attribute must be the one from `choices` list.
6. A custom data attribute of `boolean` type must not define `required` setting.

## Extracting translations

There's a custom JMS translations compatible translation extractor `ez_online_editor_attributes` which iterates over all scopes to get full list of custom attributes for all elements. It's a first step, but generated translations will require manual editing as by default attribute name is used (which is not very verbose).

Example (`dir` and `output-dir` are arbitrary and project-dependent):
```
php ./bin/console translation:extract --enable-extractor=ez_online_editor_attributes --dir=./app/Resources/views --output-dir=./app/Resources/translations/ --output-format=yaml
```

## Known issues

There is no way to set different translation per scope for the same attribute of the same element. It's element-attribute pair based.

## Naming

We can see that there are two distinct naming conventions. We already have `alloy_editor` nodes. While this configuration does not use the name directly, you can spot `OnlineEditor` in method names here. This is done on purpose. What `alloy_editor` defines is AlloyEditor-specific. However custom classes and attributes (and tags/styles for that matter) are related to Online editor features, which are independent of the editor implementation.

**TODO**:
- [x] Build basic configuration.
- [x] Implement validation of the configuration.
- [x] Implement processing of the configuration.
- [x] Define and implement scope merging rules.
- [x] Implement tests.
- [x] Expose configuration to AdminUI in ezsystems/ezplatform-admin-ui#996
- [x] Implement custom translation extractor for JMS.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
+++ b/src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
@@ -10,6 +10,7 @@ namespace EzSystems\EzPlatformRichTextBundle\DependencyInjection\Configuration\P
 
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\AbstractFieldTypeParser;
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\SiteAccessAware\ContextualizerInterface;
+use InvalidArgumentException;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
 use Symfony\Component\Config\Definition\Builder\ScalarNodeDefinition;
 
@@ -19,7 +20,22 @@ use Symfony\Component\Config\Definition\Builder\ScalarNodeDefinition;
 class RichText extends AbstractFieldTypeParser
 {
     const CLASSES_SA_SETTINGS_ID = 'fieldtypes.ezrichtext.classes';
+    const CLASSES_NODE_KEY = 'classes';
+
     const ATTRIBUTES_SA_SETTINGS_ID = 'fieldtypes.ezrichtext.attributes';
+    const ATTRIBUTES_NODE_KEY = 'attributes';
+    const ATTRIBUTE_TYPE_NODE_KEY = 'type';
+    const ATTRIBUTE_TYPE_CHOICE = 'choice';
+    const ATTRIBUTE_TYPE_BOOLEAN = 'boolean';
+    const ATTRIBUTE_TYPE_STRING = 'string';
+    const ATTRIBUTE_TYPE_NUMBER = 'number';
+
+    // constants common for OE custom classes and data attributes configuration
+    const ELEMENT_NODE_KEY = 'element';
+    const DEFAULT_VALUE_NODE_KEY = 'default_value';
+    const CHOICES_NODE_KEY = 'choices';
+    const REQUIRED_NODE_KEY = 'required';
+    const MULTIPLE_NODE_KEY = 'multiple';
 
     /**
      * Returns the fieldType identifier the config parser works for.
@@ -246,8 +262,8 @@ class RichText extends AbstractFieldTypeParser
             }
 
             $onlineEditorSettingsMap = [
-                'classes' => self::CLASSES_SA_SETTINGS_ID,
-                'attributes' => self::ATTRIBUTES_SA_SETTINGS_ID,
+                self::CLASSES_NODE_KEY => self::CLASSES_SA_SETTINGS_ID,
+                self::ATTRIBUTES_NODE_KEY => self::ATTRIBUTES_SA_SETTINGS_ID,
             ];
             foreach ($onlineEditorSettingsMap as $key => $settingsId) {
                 if (isset($scopeSettings['fieldtypes']['ezrichtext'][$key])) {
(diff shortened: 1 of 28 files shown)


bdunogier:
@alongosz remember to describe the configuration for us, outsiders, as soon as you get a chance :)

alongosz:
Provided configuration example in the description // cc @bdunogier @DominikaK

src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
```diff
@@ -18,6 +19,24 @@
  */
 class RichText extends AbstractFieldTypeParser
 {
+    const CLASSES_SA_SETTINGS_ID = 'fieldtypes.ezrichtext.classes';
```
ViniTou: nitpick: visibility for const

alongosz: Done via 189aa8b

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use EzSystems\EzPlatformRichTextBundle\DependencyInjection\Configuration\Parser\FieldType\RichText;
+use JMS\TranslationBundle\Model\MessageCatalogue;
+use JMS\TranslationBundle\Model\Message\XliffMessage;
+use JMS\TranslationBundle\Translation\ExtractorInterface;
+
+/**
+ * Generates translation strings for limitation types.
+ */
+class OnlineEditorCustomAttributesExtractor implements ExtractorInterface
```
adamwojs:
```suggestion
final class OnlineEditorCustomAttributesExtractor implements ExtractorInterface
```

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+{
+    const MESSAGE_DOMAIN = 'online_editor';
```
adamwojs:
```suggestion
    private const MESSAGE_DOMAIN = 'online_editor';
```

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+    const ATTRIBUTES_MESSAGE_ID_PREFIX = 'ezrichtext.attributes';
```
adamwojs:
```suggestion
    private const ATTRIBUTES_MESSAGE_ID_PREFIX = 'ezrichtext.attributes';
```

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+    const CLASS_LABEL_MESSAGE_ID = 'ezrichtext.classes.class.label';
```
adamwojs:
```suggestion
    private const CLASS_LABEL_MESSAGE_ID = 'ezrichtext.classes.class.label';
```

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * @var \eZ\Publish\Core\MVC\ConfigResolverInterface
+     */
+    private $configResolver;
+
+    /**
+     * @var string
```
adamwojs:
```suggestion
     * @var string[]
```

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+        foreach ($attributes as $elementName => $attributesConfig) {
+            foreach (array_keys($attributesConfig) as $attributeName) {
+                $messageId = sprintf(
+                    '%s.%s.%s.label',
+                    self::ATTRIBUTES_MESSAGE_ID_PREFIX,
+                    $elementName,
+                    $attributeName
+                );
+                $catalogue->add(
+                // by default let's use attribute name
```
adamwojs: Nitpick: Indentation is missing

src/lib/Translation/Extractor/OnlineEditorCustomAttributesExtractor.php
```diff
@@ -0,0 +1,113 @@
...
+        return $message;
+    }
+
+    /**
+     * Extract messages from the given scope into the catalogue.
+     *
+     * @param \JMS\TranslationBundle\Model\MessageCatalogue $catalogue
+     * @param string $scope
+     */
+    private function extractMessagesForScope(MessageCatalogue $catalogue, string $scope)
```
adamwojs:
```suggestion
    private function extractMessagesForScope(MessageCatalogue $catalogue, string $scope): void
```

src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
```diff
@@ -249,6 +281,13 @@ public function postMap(array $config, ContextualizerInterface $contextualizer)
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.output_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.edit_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.input_custom_xsl', $config);
+        $contextualizer->mapConfigArray(static::CLASSES_SA_SETTINGS_ID, $config);
+        // merge attributes of the same element from different scopes
+        $contextualizer->mapConfigArray(
+            static::ATTRIBUTES_SA_SETTINGS_ID,
```
adamwojs:
```suggestion
            self::ATTRIBUTES_SA_SETTINGS_ID,
```

src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
```diff
@@ -249,6 +281,13 @@ public function postMap(array $config, ContextualizerInterface $contextualizer)
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.output_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.edit_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.input_custom_xsl', $config);
+        $contextualizer->mapConfigArray(static::CLASSES_SA_SETTINGS_ID, $config);
```
adamwojs:
```suggestion
        $contextualizer->mapConfigArray(self::CLASSES_SA_SETTINGS_ID, $config);
```

alongosz:
@adamwojs applied changes you've suggested, via ac84073 and 6cfda11

SerheyDolgushev:
Just would like to highlight that `multiple` option seems to affect only the UI. And it will have no effect on the custom attribute rendering. Had this feeling after our conversation with @alongosz on slack. Otherwise, please correct me.

Example:
```
ezpublish:
    system:
        <scope>: # e.g. default or admin_group
            fieldtypes:
                ezrichtext:
                    attributes:
                        paragraph:
                            'select-single-attr':
                                type: 'choice'
                                multiple: false
                                required: true
                                choices: ['value1', 'value2', 'value3']
```
User selects `value1` and `value3` for a custom attribute using the UI. But when this custom attribute is rendered we have `value1,value3` string value.

dew326:
Yes, what would you expect?
Now in JS you can get this attribute and read the values or create a CSS selector with a combination of values (maybe not the best use case with this CSS selector but still).

alongosz:
The fix for the bug reported above was actually required on ezsystems/ezplatform-admin-ui#996 counterpart PR. Let's wait for @dew326 opinion though.

Accidentally I've discovered another bug related to normalization and fixed it via 836eb30.
I've also changed `multiple` property of custom data attributes to be `false` by default as passing multiple values was not possible in eZ Publish and it proved to be confusing for developers.

alongosz:
Reviewers: added adb8efb and f9fc501 to support custom classes and data attributes for `ezembed`, `ezembedinline`, `ezembedimage` elements and views. Changes require ezsystems/ezpublish-kernel#2665.

PR #37, opened 2019-05-07
EZP-28034: Refactored to depend on EzPlatformRestBundle

ViniTou:
> JIRA: https://jira.ez.no/browse/EZP-28034

# Desc
Extract RestBundle and Common namespace to new package.

https://github.com/ezsystems/ezpublish-kernel/pull/2630

PR #40, opened 2019-05-20
[CS] Locked php-cs-fixer at v2.15.x and aligned codebase

alongosz:
| Question           | Answer
| **JIRA issue**     | n/a
| **Bug/Improvement**| yes
| **Target version** | `1.0`, `1.1`, `master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

It seems that php-cs-fixer does not follow semver at all, so let's lock it at `2.15.x` to avoid fixing config or codebase on its every minor release.

At the same time I've decided that the new rule `php_unit_mock_short_will_return` is useful, so applied it along.

**TODO**:
- [x] Lock php-cs-fixer at `v2.15.x`
- [x] Apply CS rule `php_unit_mock_short_will_return`
- [x] Ask for Code Review.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7.0",
         "matthiasnoback/symfony-dependency-injection-test": "~3.0",
-        "friendsofphp/php-cs-fixer": "^2.13"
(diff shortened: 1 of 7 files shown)


alongosz:
@ViniTou let's take a look on this before rebasing #37

PR #44, opened 2019-06-04
EZP-30631: Fixed retrieving configuration (ezconfig) from DocBook tag

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-30631](https://jira.ez.no/browse/EZP-30631)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | latest stable
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Right now `getElementsByTagName` is used to fetch element's `ezconfig`. This kind config is used in different cases:
- custom attributes
- embed/custom tags parameters

Right now strange results sometimes are shown In cases when there are multiple nested elements with `ezconfig`. It happens because `getElementsByTagName` returns random (first?) `ezconfig` element. Instead, the first level child `ezconfig` should be returned. This PR fixes it.

1. Setup 2 custom tags: *custom_a* and *custom_b*. Both of them need to have own parameters set.
2. Start to edit any content with Rich Text field.
3. Add *custom_a* tag to any Richt Text field and set its parameters. 
4. Add *custom_b* tag Inside *custom_a*. Set some parameters for *custom_b*.
5. Save the changes.
6. Open the edited content on the front-end, and params of the nested custom tag (step 4) will be used to render the parent custom tag (step 3).

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Render.php
+++ b/src/lib/eZ/RichText/Converter/Render.php
@@ -10,6 +10,7 @@ namespace EzSystems\EzPlatformRichText\eZ\RichText\Converter;
 
 use DOMElement;
 use DOMNode;
+use DOMXPath;
 use EzSystems\EzPlatformRichText\eZ\RichText\RendererInterface;
 
 /**
@@ -37,7 +38,10 @@ abstract class Render
     protected function extractConfiguration(DOMElement $embed)
     {
         $hash = [];
-        $configElements = $embed->getElementsByTagName('ezconfig');
+
+        $xpath = new DOMXPath($embed->ownerDocument);
+        $xpath->registerNamespace('docbook', 'http://docbook.org/ns/docbook');
+        $configElements = $xpath->query('./docbook:ezconfig', $embed);
 
         if ($configElements->length) {
             $hash = $this->extractHash($configElements->item(0));
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -99,7 +99,7 @@ class Template extends Render implements Converter
         $templateType = $template->hasAttribute('type') ? $template->getAttribute('type') : 'tag';
         $parameters = [
             'name' => $templateName,
-            'params' => $this->extractTemplateConfiguration($template, $xpath),
+            'params' => $this->extractConfiguration($template),
         ];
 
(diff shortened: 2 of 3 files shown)


adamwojs:
The only thing which I'm missing in this PR is a test covering a given scenario 😉

SerheyDolgushev:
Will work on that tomorrow morning :)

SerheyDolgushev:
@adamwojs Please have a look at https://github.com/ezsystems/ezplatform-richtext/pull/44/commits/1aa39de795c29b2d7c6c7474b9451e61669b7433. This test fails without this PR.

SerheyDolgushev:
> I am not able to reproduce the problem...
> I have created tags custom_a and custom_b.
> custom_a has parameters `attr1=1` and `attr2=2`
> custom_b has parameters `attr1=3` and `attr2=4`
> I place custom_b inside custom_a
> 
> Once rendered, all attributes seems to have correct value

Per our conversation on slack: right now the "random" `ezconfig` element is fetched. Instead, the first-level child should be fetched. Otherwise, it might lead to unexpected results.

vidarl:
Yes, I misunderstood as this has to do with embeds... The embed parser calls getElementsByTagName(), and we then use the first element in the list it returns. So based on your input it is pretty random if that first node is current node, or any child.

Then indeed a xpath which selects the current node is better.

So code looks fine to me now. But possible improve PR and jira ticket description?

SerheyDolgushev:
@vidarl PR and Jira ticket description is updated per your request.

alongosz:
The PR description is not accurate. Custom attributes do not rely on ezconfig in any way. Custom Templates instead of relying on the broken method being subject of those changes, use its own method.
Moreover `ezembed>ezembed` is not supported by eZ Platform, so probably some custom validation schema was used. Therefore it's not possible for QA to test it.

That said, I appreciate the fix, so I pushed refactoring of `Template` class to rely on `extractConfiguration` to keep things DRY.

@katarzynazawada we need sanities for RichText:
- Embeds,
- Custom Tags nested inside of Custom Tags (to see if a Custom Tag uses proper configuration you can call `{{ dump(params) }}` inside Custom Tag template (or maybe just set e.g. ezyoutube with inner ezyoutube and different links - the link from outer tag should be used).

alongosz:
Thank you @SerheyDolgushev :tada:

PR #46, opened 2019-06-13
EZP-30393: Moved all Online Editor front-end code to the RichText Bundle

dew326:
| Question           | Answer
| **JIRA issue**     | [EZP-30393](https://jira.ez.no/browse/EZP-30393)
| **Requires:** | <ul><li>ezsystems/ezplatform-page-builder#354</li><li>ezsystems/ezplatform-admin-ui#1014</li></ul>
| **Improvement**| yes
| **New feature**    | no
| **Target version** | `master (2.0@dev)` for eZ Platform `v3.0`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | yes

### Summary

This PR moves front-end code and back-end configuration related to AlloyEditor from AdminUI, keeping there minimal code just for embedding RichText.

### Backward incompatible changes

* `eZ.richText` with proper keys replaced `eZ.adminUiConfig.alloyEditor`, `eZ.adminUiConfig.richTextCustomTags`, and `eZ.adminUiConfig.richTextCustomStyles`,
* All assets (js, scss, svg, fonts, etc.) have been moved from AdminUI bundle here, custom 3rd party development around OE needs to align with that.

### New features

* Configuration for front-end is exposed as `eZ.richText`.
* Configuration can be also used in Twig templates using The `ez_richtext_config` Twig function.
* Configuration is now exposed via PHP API by means of auto-wirable `\EzSystems\EzPlatformRichText\API\Configuration\ProviderService` Symfony service.
* Auto-configurable Service Provider Interface `\EzSystems\EzPlatformRichText\SPI\Configuration\Provider` allows extending configuration by custom settings.

### Documentation

eZ Platform RichText Bundle exposes Twig global variable `ez_richtext_config` which contains RichText and AlloyEditor configuration. In AdminUI this configuration is exposed for front-end as `eZ.richText` JavaScript variable and currently out-of-the-box provides the following configs: `alloyEditor`, `customStyles`, `customTags`. Please note that in any case AdminUI requires the RichText Bundle to be enabled.

The bundle also provides PHP API to retrieve that configuration. To use it, inject `\EzSystems\EzPlatformRichText\API\Configuration\ProviderService` (the service supports autowiring).

It is also possible to extend the configuration by implementing the SPI `\EzSystems\EzPlatformRichText\SPI\Configuration\Provider`. It's enough to register service implementing that interface and autoconfigure it.

**TODO**:
- [x] Move custom attribute when it's merged (https://github.com/ezsystems/ezplatform-admin-ui/pull/996).
- [x] Move AE list fix (https://github.com/ezsystems/ezplatform-admin-ui/pull/1015)
- [x] Move configuration (backend)
      - Move `eZ.adminUiConfig.alloyEditor` to `eZ.richText`,
      - Move `eZ.adminUiConfig.richTextCustomTags` to `eZ.richText.customTags`,
      - Move `eZ.adminUiConfig.richTextCustomStyles` to `eZ.richText.customStyles`.


src/bundle/Resources/config/ui/mappers.yaml
```diff
@@ -0,0 +1,41 @@
...
+        public: false
+
+    # RichText Custom Tags UI config attribute type mappers
+    EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomTag\CommonAttributeMapper:
+        tags:
+            - { name: ezrichtext.configuration.custom_tag.mapper, priority: 0 }
+
+    EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomTag\ChoiceAttributeMapper:
+        parent: EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomTag\CommonAttributeMapper
+        autowire: true
```
adamwojs:
```suggestion
```

`true` is the default value for `autowire` option in this file.

src/lib/Configuration/AggregateProvider.php
```diff
@@ -0,0 +1,40 @@
...
+namespace EzSystems\EzPlatformRichText\Configuration;
+
+use EzSystems\EzPlatformRichText\API\Configuration\ProviderService;
+
+/**
+ * RichText configuration provider, providing configuration by aggregating different sources.
+ *
+ * @internal type-hint \EzSystems\EzPlatformRichText\API\Configuration\ProviderService
+ */
+final class AggregateProvider implements ProviderService
```
adamwojs: Unit tests are missing for this class.

src/bundle/Templating/Twig/Extension/RichTextConfigurationExtension.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /** @var \EzSystems\EzPlatformRichText\API\Configuration\ProviderService */
+    private $configurationProvider;
+
+    public function __construct(Configuration\ProviderService $configurationProvider)
+    {
+        $this->configurationProvider = $configurationProvider;
+    }
+
+    public function getName()
```
mikadamczyk:
```suggestion
    public function getName(): string
```

src/lib/Configuration/Provider/AlloyEditor.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @deprecated 3.0.0 The alternative and more flexible solution will be introduced.
+     * @deprecated 3.0.0 So you will need to update Online Editor Extra Buttons as part of eZ Platform 3.x upgrade.
+     *
+     * @return array Custom buttons
+     */
+    private function getExtraButtons(): array
+    {
+        @trigger_error(
+            '"ezrichtext.alloy_editor.extra_buttons" is deprecated since v2.5.1. There will be new and more flexible solution to manage buttons in Online Editor in 3.0.0',
```
mikadamczyk: can you split this into lines to make it more readable?

src/lib/Configuration/Provider/CustomStyle.php
```diff
@@ -0,0 +1,58 @@
...
+final class CustomStyle implements Provider
+{
+    /** @var \eZ\Publish\Core\MVC\ConfigResolverInterface */
+    private $configResolver;
+
+    /** @var \EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomStyle */
+    private $customStyleConfigurationMapper;
+
+    /**
+     * @param \eZ\Publish\Core\MVC\ConfigResolverInterface $configResolver
```
mikadamczyk:
```suggestion
```

src/lib/Configuration/Provider/CustomStyle.php
```diff
@@ -0,0 +1,58 @@
...
+     * @param \EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomStyle $customStyleConfigurationMapper
```
mikadamczyk:
```suggestion

```

src/lib/Configuration/Provider/CustomTag.php
```diff
@@ -0,0 +1,58 @@
...
+final class CustomTag implements Provider
+{
+    /** @var \eZ\Publish\Core\MVC\ConfigResolverInterface */
+    private $configResolver;
+
+    /** @var \EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomTag */
+    private $customTagConfigurationMapper;
+
+    /**
+     * @param \eZ\Publish\Core\MVC\ConfigResolverInterface $configResolver
```
mikadamczyk: Those are not needed

src/lib/Configuration/UI/Mapper/CustomTag/ChoiceAttributeMapper.php
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Map RichText Custom Tag attribute of 'choice' type to proper UI config.
+ *
+ * @internal For internal use by RichText package
+ */
+final class ChoiceAttributeMapper extends CommonAttributeMapper implements AttributeMapper
+{
+    /**
+     * {@inheritdoc}
```
mikadamczyk: Should we still use `{@inheritdoc}`?

src/lib/Configuration/UI/Mapper/OnlineEditor.php
```diff
@@ -0,0 +1,93 @@
...
+final class OnlineEditor implements OnlineEditorConfigMapper
+{
+    /** @var \Symfony\Component\Translation\TranslatorInterface */
+    private $translator;
+
+    /** @var string */
+    private $translationDomain;
+
+    /**
+     * @param \Symfony\Contracts\Translation\TranslatorInterface $translator
```
mikadamczyk:
```suggestion

```

src/lib/Configuration/UI/Mapper/OnlineEditor.php
```diff
@@ -0,0 +1,93 @@
...
+     * @param string $translationDomain
```
mikadamczyk:
```suggestion
```

tests/lib/Configuration/UI/Mapper/OnlineEditorTest.php
```diff
@@ -0,0 +1,217 @@
...
+    private $mapper;
+
+    public function setUp(): void
+    {
+        $translatorMock = $this->createMock(TranslatorInterface::class);
+        $translatorMock
+            ->expects($this->any())
+            ->method('trans')
+            ->willReturnArgument(0);
+        /** @var \Symfony\Contracts\Translation\TranslatorInterface $translatorMock */
```
mikadamczyk: This could be move where mock is created

src/lib/Configuration/AggregateProvider.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace EzSystems\EzPlatformRichText\Configuration;
+
+use EzSystems\EzPlatformRichText\API\Configuration\ProviderService;
+
+/**
+ * RichText configuration provider, providing configuration by aggregating different sources.
+ *
+ * @internal type-hint \EzSystems\EzPlatformRichText\API\Configuration\ProviderService
```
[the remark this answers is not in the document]
alongosz: @adamwojs this one is intentional. We use the phrase "for internal use..." if there's no better explanation. However here most common mistake would be to refer to the implementation instead of API interface. This doc explains to a developer how to resolve that (you can see it if you hover over struck-through definition in your code).

src/bundle/Resources/config/ui/mappers.yaml
alongosz: > `true` is the default value for `autowire` option in this file.

```
Attribute "autowire" on service "EzSystems\EzPlatformRichText\Configuration\UI\Mapper\CustomTag\ChoiceAttributeMapper" cannot be inherited from "_defaults" when a "parent" is set. Move   
  your child definitions to a separate file or define this attribute explicitly
```
(the former one seems to be an overkill here)

src/lib/Configuration/AggregateProvider.php
alongosz: Added via 00f92787c.

I've also added missing coverage for configuration providers which are passed to that aggregate. However I'm not really sure about them. It's mock on top of mock to describe control flow instead of something more essential (see 47a66d6e7 and 99bf08ed2). We need here some Symfony full stack end-to-end test which for given semantic config will return expected UI configuration.

Creating those tests forced me to create common interface for `CustomStyle` and `CustomTag` providers to mock them, which, given the nature of them - stored as custom Templates - is probably a good thing... well, maybe except for auto-wiring no longer working. See d8cb3e0eb.

alongosz:
Addressed all CR issues, please review:
- 4dc9b6df6
- 7458d3154
- 31ecd7d2f
- bddd421fd
- 6acd3d112
- 00f92787c
- 47a66d6e7
- d8cb3e0eb
- 99bf08ed2

src/lib/Configuration/Provider/AlloyEditor.php
alongosz: Actually, I've decided to modify behavior here a bit as well. Till now it reported that warning on every load of AdminUI, even if user didn't set that configuration. So I've changed it to check if it's used in the first place. See 6acd3d112. This will reduce size of deprecation log.

composer.json
```diff
@@ -18,6 +18,8 @@
         "symfony/options-resolver": "^4.3",
         "symfony/validator": "^4.3",
         "symfony/cache": "^4.3",
+        "symfony/translation": "^4.3",
+        "symfony/translation-contracts": "^1.1.2",
```
webhdx: Why not `^1.1`?

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -25,6 +26,7 @@ class EzPlatformRichTextExtension extends Extension implements PrependExtensionI
     const RICHTEXT_CUSTOM_STYLES_PARAMETER = 'ezplatform.ezrichtext.custom_styles';
     const RICHTEXT_CUSTOM_TAGS_PARAMETER = 'ezplatform.ezrichtext.custom_tags';
     const RICHTEXT_ALLOY_EDITOR_PARAMETER = 'ezplatform.ezrichtext.alloy_editor';
+    public const RICHTEXT_CONFIGURATION_PROVIDER_TAG = 'ezrichtext.configuration.provider';
```
webhdx: I suppose we prefix all service tags with either `ezplatform` or `ezpublish`?

src/bundle/Resources/config/configuration.yaml
```diff
@@ -0,0 +1,31 @@
...
+    # set via compiler pass
+    ezplatform.ezrichtext.alloy_editor: []
+
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    EzSystems\EzPlatformRichText\Configuration\Provider\:
```
webhdx: I'd remove this definition as it's too greedy. Also you are registering explicitly all classes under `EzSystems\EzPlatformRichText\Configuration\Provider` namespace below.

src/bundle/Templating/Twig/Extension/RichTextConfigurationExtension.php
```diff
@@ -0,0 +1,41 @@
...
+    private $configurationProvider;
+
+    public function __construct(Configuration\ProviderService $configurationProvider)
+    {
+        $this->configurationProvider = $configurationProvider;
+    }
+
+    public function getName(): string
+    {
+        return 'ezpublish.rich_text';
```
webhdx: I've noticed it's called `richtext`, `ezrichtext` or `rich_text`. Since it's targeted at 3.0 release, can't we streamline it?

composer.json
alongosz: > Why not `^1.1`?

Because of composer memory issues. It's better to include latest possible version. Especially now that we don't need to wonder if that breaks someone's project. After 3.0 final we'd need to be more careful, maybe.

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
alongosz: > I suppose we prefix all service tags with either `ezplatform` or `ezpublish`?

Ok, fixed prefixed via `ezplatform` 45b1cba1eab314f997869860c8ec55a65ab815bd. We're not going to add `ezpublish` to any new code.
Also fixed remaining service tag names via #65

src/bundle/Resources/config/configuration.yaml
alongosz: > I'd remove this definition as it's too greedy. Also you are registering explicitly all classes under `EzSystems\EzPlatformRichText\Configuration\Provider` namespace below.

Right, I thought AlloyEditor was still configured by this, but I see now that it's explicit as well. Fixed via b333cb1fdd31730e6d084370ca570f98ad593879

src/bundle/Templating/Twig/Extension/RichTextConfigurationExtension.php
alongosz: > I've noticed it's called `richtext`, `ezrichtext` or `rich_text`. Since it's targeted at 3.0 release, can't we streamline it?

Good catch, I must have missed this when copying it from somewhere else.
Current prefix for Twig extensions is `ezrichtext` so I've changed this to `ezrichtext.configuration` via 7f25704a2e9319ea3dcedf0f7695d2fa1222d728. It's in the form of `"ezrichtext"` here to follow Symfony bundle extension name.

We can add `ezplatform` prefix as well, but the other extensions needs alignment in this case.

alongosz:
> Good work, just a few comments from my side. What I find lacking is a description on what was added. I was not sure if I should write comments on the code which was most likely just moved and not even belong to you.

@webhdx I've addressed all your comments. I value your input regardless of if this is a copied code or a new one. All the reviews for this PR prove that reviewing copied code is also useful.
That said, I've updated the description as well so it's more clear.

PR #48, opened 2019-06-26
EZP-30628: Soft return inside inline style is removed after edit

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-30628](https://jira.ez.no/browse/EZP-30628)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 1.1
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Soft return placed in the text wrapped with custom style isn't preserved after switching to edit mode. Related to https://github.com/ezsystems/ezplatform-richtext/pull/34 (proper rendering in view mode).

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
@@ -729,7 +729,7 @@
     <xsl:apply-templates select="node()|@*"/>
   </xsl:template>
 
-  <xsl:template match="docbook:eztemplateinline/docbook:ezcontent/text()">
+  <xsl:template match="docbook:literallayout/docbook:eztemplateinline/docbook:ezcontent/text()">
     <xsl:call-template name="breakLine">
       <xsl:with-param name="text" select="."/>
     </xsl:call-template>
--- a/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
@@ -729,6 +729,12 @@
     <xsl:apply-templates select="node()|@*"/>
   </xsl:template>
 
+  <xsl:template match="docbook:eztemplateinline/docbook:ezcontent/text()">
+    <xsl:call-template name="breakLine">
+      <xsl:with-param name="text" select="."/>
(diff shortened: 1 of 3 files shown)


alongosz:
This will probably cause issues for migrations and our demo installation, as #34 did. While you might want soft returns (CRLF or just LF) to be transformed to breaks when editing and submitting forms, when preparing migration, you tend to format XML correctly, which then adds some unexpected `br`.

I'd prefer to handle it via DocBook tag. White-spaces are not part of XML spec and it was my mistake to accept #34

But maybe this one won't be a problem for us? POV ping @damianz5 @ciastektk

adamwojs:
@alongosz Which exactly DocBook tag you have in mind?

alongosz:
> @alongosz Which exactly DocBook tag you have in mind?

And that's something I need someone to figure out. Is `br` part of DocBook as well? Of course if there's no other way, I'll accept it. But the current solution causes problems. We would also need to consult with @dew326 if it's possible to listen to shift+enter in OE and replace it with tag. If it's not, then there's no point in discussing it further.

dew326:
AlloyEditor makes `br` on shift+enter and expects the same in the input in edit mode.

adamwojs:
> And that's something I need someone to figure out. Is br part of DocBook as well? 

Unfortunately `br` is not part of the DocBook specification and AFAIK there is not equivalent tag. 

> We would also need to consult with @dew326 if it's possible to listen to shift+enter in OE and replace it with tag.

AFAICS It's already inserting `<br />` when you press `shift+enter` 😉

alongosz:
We might need to extend eZ Platform DocBook to support this tag. Right now `<br>` inside `<p>` results in `<para><literallayout>\n` which is weird.

konradoboza:
As far as I can tell, `<para>` could be parent for `<literallayout>`, ref: https://tdg.docbook.org/tdg/5.2/literallayout.html, section "Parents".

alongosz:
> As far as I can tell, `<para>` could be parent for `<literallayout>`, ref: https://tdg.docbook.org/tdg/5.2/literallayout.html, section "Parents".

It doesn't matter that it could. You _can_ also set with of an element with JS instead of CSS, but is it a correct solution?

lserwatka:
You can merge it up

konradoboza:
Done.

PR #54, opened 2019-07-31
EZP-30730: Narrowed "LF to <br>" XSLT to literallayout scope

alongosz:
| Question           | Answer
| **JIRA issue**     | [EZP-30730](https://jira.ez.no/browse/EZP-30730)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `v1.1.4` for eZ Platform `v2.5.4 LTS`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

This PR narrows down the scope of applied transformation rendering HTML `<br>` tags for nested Custom Templates (Tags/Styles) content to `docbook:literallayout` contents.

The issue is that at the moment of applying transformation of `docbook:ezcontent`, for the sample Custom Style:
```xml
  <eztemplate name="style3" type="style">
    <ezcontent>Line
break
    </ezcontent>
  </eztemplate>
```
the processed on the fly XML has the form of:
```xml
<?xml version="1.0"?>\n
<section xmlns="http://docbook.org/ns/docbook">Line
break
</section>
```

The solution is to wrap block Custom Styles with `literallayout` when needed.

To achieve that the transformation from HTML5Edit transforms DocBook differently:

Before
```xml
<eztemplate name="style3" type="style">
    <ezcontent>Line
break
    </ezcontent>
```
After:
```xml
<eztemplate name="style3" type="style">
    <ezcontent><literallayout>Line
break</literallayout>
    </ezcontent>
```

It's needed for block Custom Styles only, because inline Custom Templates are already wrapped by transformations outer block elements:
```xml
<para>
    <literallayout>
        <eztemplateinline>
            ...
        </eztemplateinline>
    </literallayout>
</para>
```

Moreover block Custom Tags allow inserting block elements only, so specific block element transformation is already applied, as in the case above.

### BC

For backward compatibility reasons the following stored XML:
```xml
<eztemplate name="style3" type="style">
    <ezcontent>Line
break
    </ezcontent>
```
needs to produce proper line breaks for the output as well.

Given that the previous implementation applied line breaking transformation to `section/text()`, entire `ezcontent` is wrapped with `literallayout` tag, unless it explicitly contains that tag.

### Background

Providing this verbatim payload as RichText Field Value `DOMDocument`:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<section xmlns="http://docbook.org/ns/docbook" xmlns:xlink="http://www.w3.org/1999/xlink"
         xmlns:ezxhtml="http://ez.no/xmlns/ezpublish/docbook/xhtml"
         xmlns:ezcustom="http://ez.no/xmlns/ezpublish/docbook/custom"
         version="5.0-variant ezpublish-1.0">
  <para>
Text
  </para>
```
results in the following output HTML:
```html
<br>
<p>
<br>
Text
<br>
</p>
<br>
```
This is a side effect of CSR-required change introduced via #34 for Custom Templates due to unavailability of outer scope when processing `ezcontent` of a Template (as described in the previous section).

The eZ Platform DocBook spec defines `literallayout` as the only tag where `breakline` transformation should be applied, so current behavior is incorrect. Wrapping literal `ezcontent` with `literallayout` solves that challenge as applying line breaking transformation on `section/literallayout` is correct with respect to the specification.

### QA

The bug is visible with eZ Platform EE Demo `2.5` on the landing page inside block of "Top Place to Go". It renders `<br>` which makes content a little bit misaligned.

Please test against regressions to [EZP-29876](https://jira.ez.no/browse/EZP-29876) and [EZP-30628](https://jira.ez.no/browse/EZP-30628).

Please also test BC by storing RichText according to those JIRA tickets, and then checking out changes from this PR to see if w/o editing line breaks still exist. Please then see if editing is still possible.

Line breaks are applied to the following block and inline elements:
- paragraph,
- emphasis,
- table cell,
- Custom template (Style or Tag).

**TODO**:
- [x] Provide better PR description.
- [x] Narrow `LF` to `<br>` XSL transformation to Custom Tags/Styles inside `literallayout` only.
- [x] Provide BC for stored DocBook with Custom Styles.
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -23,6 +23,8 @@ use Psr\Log\NullLogger;
  */
 class Template extends Render implements Converter
 {
+    const LITERAL_LAYOUT_LINE_BREAK = "\n";
+
     /**
      * @var \EzSystems\EzPlatformRichText\eZ\RichText\Converter
      */
@@ -235,7 +237,7 @@ class Template extends Render implements Converter
      */
     private function wrapContentWithLiteralLayout(DOMNode $rootNode, DOMNode $node): DOMNode
     {
-        if (false === strpos($node->nodeValue, "\n")) {
+        if (false === strpos($node->nodeValue, self::LITERAL_LAYOUT_LINE_BREAK)) {
             return $rootNode;
         }
 
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -235,20 +235,22 @@ class Template extends Render implements Converter
      */
     private function wrapContentWithLiteralLayout(DOMNode $rootNode, DOMNode $node): DOMNode
     {
-        $xpath = new DOMXPath($node->ownerDocument);
-        $xpath->registerNamespace('docbook', 'http://docbook.org/ns/docbook');
-        if (
-            strpos($node->nodeValue, "\n") !== false
-            && $xpath->query('.//docbook:literallayout', $node)->length === 0
-        ) {
-            $literalLayoutNode = $rootNode->ownerDocument->createElementNS(
-                'http://docbook.org/ns/docbook',
-                'literallayout'
-            );
-
-            return $rootNode->appendChild($literalLayoutNode);
+        if (false === strpos($node->nodeValue, "\n")) {
+            return $rootNode;
         }
 
-        return $rootNode;
+        $xpath = new DOMXPath($node->ownerDocument);
+        $xpath->registerNamespace('docbook', 'http://docbook.org/ns/docbook');
+
+        if ($xpath->query('.//docbook:literallayout', $node)->length > 0) {
+            return $rootNode;
+        }
+
+        $literalLayoutNode = $rootNode->ownerDocument->createElementNS(
+            'http://docbook.org/ns/docbook',
+            'literallayout'
+        );
+
+        return $rootNode->appendChild($literalLayoutNode);
     }
 }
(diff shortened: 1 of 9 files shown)


src/lib/eZ/RichText/Converter/Render/Template.php
```diff
@@ -221,4 +223,32 @@ protected function extractTemplateConfiguration(DOMElement $template, DOMXPath $
...
+     * @param \DOMNode $node
+     *
+     * @return \DOMNode
+     */
+    private function wrapContentWithLiteralLayout(DOMNode $rootNode, DOMNode $node): DOMNode
+    {
+        $xpath = new DOMXPath($node->ownerDocument);
+        $xpath->registerNamespace('docbook', 'http://docbook.org/ns/docbook');
+        if (
+            strpos($node->nodeValue, "\n") !== false
```
adamwojs:
```suggestion
            strpos($node->nodeValue, PHP_EOL) !== false
```

alongosz: > ```diff
> +          strpos($node->nodeValue, "\n") !== false
> -          strpos($node->nodeValue, PHP_EOL) !== false
> ```
I would disagree with that one. This is not `PHP_EOL`, this is [DocBook "\n"](https://github.com/ezsystems/ezplatform-richtext/blob/6cdb5054c6ee2eda28d0f10a8a2b45b2f6b50a89/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl#L93) which accidentally is the same.
Maybe it calls for dedicated const, not sure however in which class to put it.

alongosz:
Applied @mikadamczyk and @adamwojs suggestions after internal sync: ca358271d, ea3696957.

lserwatka:
@adamwojs could you merge it up here?

adamwojs:
@lserwatka Done.

PR #65, opened 2019-08-26
Prefixed Service Tags with ezplatform to align with existing convention

alongosz:
| Question           | Answer
| **JIRA issue**     | n/a
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `master (2.0@dev)` for eZ Platform `v3.0`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | yes

This PR adds `ezplatform.` prefix to all service tags provided by the RichText bundle to keep consistency with the other eZ Platform bundles.

### Open question

There are services defined by old naming scheme (instead of FQCN). All the names start with `ezrichtext.` as well. Should we also prefix them with `ezplatform.`?

Should we prefix in the same way parameters and Twig extension names?

### Documentation

Release notes and/or upgrade instructions should mention renamed service tags.

**TODO**:
- [ ] Merge #46 first.
- [x] Rename the service tag `ezrichtext.converter.output.xhtml5` to `ezplatform.ezrichtext.converter.output.xhtml5`.
- [x] Rename the service tag `ezrichtext.converter.input.xhtml5` to `ezplatform.ezrichtext.converter.input.xhtml5`.
- [x] Rename the service tag `ezrichtext.validator.input.ezxhtml5` to `ezplatform.ezrichtext.validator.input.ezxhtml5`.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/DependencyInjection/Compiler/RichTextHtml5ConverterPass.php
+++ b/src/bundle/DependencyInjection/Compiler/RichTextHtml5ConverterPass.php
@@ -24,13 +24,13 @@ class RichTextHtml5ConverterPass implements CompilerPassInterface
     {
         if ($container->hasDefinition('ezrichtext.converter.output.xhtml5')) {
             $html5OutputConverterDefinition = $container->getDefinition('ezrichtext.converter.output.xhtml5');
-            $taggedOutputServiceIds = $container->findTaggedServiceIds('ezrichtext.converter.output.xhtml5');
+            $taggedOutputServiceIds = $container->findTaggedServiceIds('ezplatform.ezrichtext.converter.output.xhtml5');
             $this->setConverterDefinitions($taggedOutputServiceIds, $html5OutputConverterDefinition);
         }
 
         if ($container->hasDefinition('ezrichtext.converter.input.xhtml5')) {
             $html5InputConverterDefinition = $container->getDefinition('ezrichtext.converter.input.xhtml5');
-            $taggedInputServiceIds = $container->findTaggedServiceIds('ezrichtext.converter.input.xhtml5');
+            $taggedInputServiceIds = $container->findTaggedServiceIds('ezplatform.ezrichtext.converter.input.xhtml5');
             $this->setConverterDefinitions($taggedInputServiceIds, $html5InputConverterDefinition);
         }
     }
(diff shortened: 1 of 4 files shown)


pawbuj (approved):

src/bundle/DependencyInjection/Compiler/RichTextHtml5ConverterPass.php
```diff
@@ -24,13 +24,13 @@ public function process(ContainerBuilder $container)
     {
         if ($container->hasDefinition('ezrichtext.converter.output.xhtml5')) {
             $html5OutputConverterDefinition = $container->getDefinition('ezrichtext.converter.output.xhtml5');
-            $taggedOutputServiceIds = $container->findTaggedServiceIds('ezrichtext.converter.output.xhtml5');
+            $taggedOutputServiceIds = $container->findTaggedServiceIds('ezplatform.ezrichtext.converter.output.xhtml5');
```
adamwojs: Should we introduce BC layer for the old `ezrichtext.converter.output.xhtml5` tag? I mean something similar to what we did in case of field type related tags? https://github.com/ezsystems/ezpublish-kernel/blob/94214dfc4c6d3af87714e1eb3b800972f7b7a18e/eZ/Publish/Core/Base/Container/Compiler/AbstractFieldTypeBasedPass.php#L36-L46

alongosz: > Should we introduce BC layer for the old `ezrichtext.converter.output.xhtml5` tag? I mean something similar to what we did in case of field type related tags? https://github.com/ezsystems/ezpublish-kernel/blob/94214dfc4c6d3af87714e1eb3b800972f7b7a18e/eZ/Publish/Core/Base/Container/Compiler/AbstractFieldTypeBasedPass.php#L36-L46

I wondered the same myself. This service tag was named like that since eZ Platform 2.4, so per OSS BC best practices we should. However we've decided that for 3.0 we don't want code polluted with deprecation.
Let me suggest a compromise. Maybe I'll re-target this to `1.1` (corresponding to the next eZ Platform `2.5.x LTS`) and add it there in BC safe way with deprecation? Then, for `master` I would drop old service tag.

PR #76, opened 2019-10-24
EZP-31076: Marking embed content as hidden doesn't change its visibility

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-31076](https://jira.ez.no/browse/EZP-31076)
| **Bug**| yes
| **New feature**    | no
| **Target version** | `1.1`/`master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

As mentioned in the JIRA ticket, currently embedded content is displayed even if it's hidden. This PR fixes this incorrect behavior. 

**TODO**:
- [X] Implement tests.
- [X] Ask for Code Review.

### DIFF
--- a/src/bundle/eZ/RichText/Renderer.php
+++ b/src/bundle/eZ/RichText/Renderer.php
@@ -155,6 +155,14 @@ class Renderer implements RendererInterface
                 return null;
             }
 
+            if ($content->contentInfo->isHidden) {
+                $this->logger->error(
+                    "Could not render embedded resource: Content #{$contentId} is hidden."
+                );
+
+                return null;
+            }
+
             $this->checkContentPermissions($content);
         } catch (AccessDeniedException $e) {
             $this->logger->error(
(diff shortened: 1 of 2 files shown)


kmadejski:
@andrerom it seems that it's not necessary once `Content` is hidden, all of it's `Locations` are `invisible`. This case is already handled in https://github.com/ezsystems/ezplatform-richtext/blob/1.1/src/bundle/eZ/RichText/Renderer.php#L212.

@konradoboza I'll push it to the branch in this repo. Back in times, it was a good trick for Travis.

kmadejski:
Reopened in https://github.com/ezsystems/ezplatform-richtext/pull/77 from this repo's branch.

bdunogier:
My main concern is that this is a behaviour change. By upgrading to the version that has this feature, items that were visible are suddenly not visible. Could we make this an opt-in change in 2.x, and the default in 3.0 ?

PR #77, opened 2019-10-24
EZP-31076: Skipped rendering hidden embedded content

kmadejski:
| Question           | Answer
| **JIRA issue**     | [EZP-31076](https://jira.ez.no/browse/EZP-31076)
| **Bug**| yes
| **New feature**    | no
| **Target version** | `1.1`/`master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

As mentioned in the JIRA ticket, currently embedded content is displayed even if it's hidden. This PR fixes this incorrect behavior. 

**TODO**:
- [X] Implement tests.
- [X] Ask for Code Review.

### DIFF
--- a/src/bundle/eZ/RichText/Renderer.php
+++ b/src/bundle/eZ/RichText/Renderer.php
@@ -155,6 +155,14 @@ class Renderer implements RendererInterface
                 return null;
             }
 
+            if ($content->contentInfo->isHidden) {
+                $this->logger->error(
+                    "Could not render embedded resource: Content #{$contentId} is hidden."
+                );
+
+                return null;
+            }
+
             $this->checkContentPermissions($content);
         } catch (AccessDeniedException $e) {
             $this->logger->error(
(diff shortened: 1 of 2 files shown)


bdunogier:
My main concern is that this is a behaviour change. By upgrading to the version that has this feature, items that were visible are suddenly not visible. Could we make this an opt-in change in 2.x, and the default in 3.0 ?

kmadejski:
> My main concern is that this is a behaviour change. By upgrading to the version that has this feature, items that were visible are suddenly not visible.

@bdunogier Thanks for sharing your point of view. Unfortunately, I disagree. The fact that the content was visible is just a clear bug. If you think about that in a very strict manner, you could even say that this is a security issue, since the content which you expect to be hidden suddenly becomes visible and you have no control over this behavior at all.

> Could we make this an opt-in change in 2.x, and the default in 3.0 ?

Technically, yes, we can, but I would like to keep it as it is now.

lserwatka:
Could you merge it up?

kmadejski:
@lserwatka done in https://github.com/ezsystems/ezplatform-richtext/commit/a81dc2f5d63609ccab15fa21d3b657fecf57ba68.

PR #84, opened 2019-11-13
EZP-31130: Ensured content variable is always present in Custom Template Twig view

alongosz:
| Question           | Answer
| **JIRA issue**     | [EZP-31130](https://jira.ez.no/browse/EZP-31130)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `v1.1.x` for eZ Platform `2.5.x` LTS
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no, doc is already aligned

This PR introduces small change - ensuring that `content` Twig variable is always present in Custom Template (Tag, Style) Twig view.

eZ Platform DocBook schema (`.rng`) currently allows `ezcontent` tag to be not present in `eztemplate`. While AdminUI ensures `ezcontent` is always set, even if empty, a DocBook comming from migrated `XmlText` might not have this tag sometimes. This resulted in `content` Twig variable being not present in a Twig view for a given Custom Template, which yields a fatal error both on a front site and in AdminUI Location View.

There were 3 directions to solve that issue:
1. Force a developer to check for existence of that variable.
2. Add `ezcontent` tag during migration from XmlText to RichText.
3. Make `content` variable always present in a Twig view for Custom Template.

The second option was rejected because RNG schema defines `ezcontent` as optional.
The first option yields bad DX.

Currently this PR implements the third option, so a developer, instead of writing:
``` twig
{% if content is defined %}{{ content|raw }}{% endif %}
```
can simply write:
``` twig
{{ content|raw }}
```
which is also [consistent with the doc that we provide](https://doc.ezplatform.com/en/2.5/guide/extending/extending_online_editor/#rendering).

### QA

- [ ] Sanity checks for creating Custom Tags with or without content.
- [ ] Test: Create a Custom Tag, find it in `ezcontentobject_attributes`, remove `<content>...</content>` markup from XML, save, clear cache, observe the mentioned error on a front site.

**TODO**:
- [x] Always expose `content` variable in Twig views for Custom Templates.
- [x] Align tests.
- [ ] Rebase after merging #83
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -107,9 +107,7 @@ class Template extends Render implements Converter
         foreach ($contentNodes as $contentNode) {
             $innerContent .= $this->getCustomTemplateContent($contentNode);
         }
-        if (!empty($innerContent)) {
-            $parameters['content'] = $innerContent;
-        }
+        $parameters['content'] = !empty($innerContent) ? $innerContent : null;
 
         if ($template->hasAttribute('ezxhtml:align')) {
             $parameters['align'] = $template->getAttribute('ezxhtml:align');
--- a/tests/lib/eZ/RichText/Converter/Render/TemplateTest.php
+++ b/tests/lib/eZ/RichText/Converter/Render/TemplateTest.php
@@ -8,11 +8,11 @@ declare(strict_types=1);
 
 namespace EzSystems\Tests\EzPlatformRichText\eZ\RichText\Converter\Render;
 
-use PHPUnit\Framework\TestCase;
+use DOMDocument;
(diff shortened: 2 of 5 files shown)


vidarl:
FYI : I just realized that AdminUI does NOT ensure `<ezcontent/>`is set on inline custom tags if they have no content

alongosz:
> FYI : I just realized that AdminUI does NOT ensure `<ezcontent/>`is set on inline custom tags if they have no content

Hmm, ok. Just one more reason for the change proposed here :+1:

lserwatka:
You can merge it up.

alongosz:
> You can merge it up.

Done via 0b1045f31.

PR #87, opened 2019-11-29
EZP-31118: Introduced strict types for ContenteService

adamwojs:
| Question           | Answer
| **JIRA issue**     | [EZP-31118](https://jira.ez.no/browse/EZP-31118)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** |  `master`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | yes

Fixed issues reported in ezsystems/ezpublish-kernel#2866 (review):

```
Argument 1 passed to eZ\Publish\Core\Repository\ContentService::loadContent() must be of the type int, string given, called in /home/vagrant/master/vendor/ezsystems/ezplatform-richtext/src/bundle/eZ/RichText/Renderer.php on line 130
```

**TODO**:
- [ ] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/eZ/RichText/Renderer.php
+++ b/src/bundle/eZ/RichText/Renderer.php
@@ -127,7 +127,7 @@ class Renderer implements RendererInterface
             /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
             $content = $this->repository->sudo(
                 function (Repository $repository) use ($contentId) {
-                    return $repository->getContentService()->loadContent($contentId);
+                    return $repository->getContentService()->loadContent((int)$contentId);
                 }
             );
 


src/bundle/eZ/RichText/Renderer.php
```diff
@@ -127,7 +127,7 @@ public function renderContentEmbed($contentId, $viewType, array $parameters, $is
             /** @var \eZ\Publish\API\Repository\Values\Content\Content $content */
             $content = $this->repository->sudo(
                 function (Repository $repository) use ($contentId) {
-                    return $repository->getContentService()->loadContent($contentId);
+                    return $repository->getContentService()->loadContent((int)$contentId);
```
alongosz: TBH you should fix it [here](https://github.com/ezsystems/ezplatform-richtext/blob/v2.0.0-beta4/src/lib/eZ/RichText/Converter/Render/Embed.php#L89) and enforce strict types in `renderContentEmbed`.

But can be done as a follow-up when handling `renderContentEmbed` by introducing strict types there and removing type cast here.

PR #93, opened 2019-12-19
EZP-30129: Handled bold & italic styles when converting from OE format to DocBook

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-30129](https://jira.ez.no/browse/EZP-30129)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 1.1 (2.5)
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Format coming from copy /paste will often use `<b>` and `<i>` tags, or sometimes also span tags with bold / italic style attribute. This change makes sure to map these cases to the correct docbook tags we support. Solving the linked Word copy / paste issue, as well as improving the code for import use cases from other systems.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -140,7 +140,7 @@
     </blockquote>
   </xsl:template>
 
-  <xsl:template match="ezxhtml5:em">
+  <xsl:template match="ezxhtml5:em | ezxhtml5:i">
     <emphasis>
       <xsl:if test="@class">
         <xsl:attribute name="ezxhtml:class">
@@ -154,7 +154,7 @@
     </emphasis>
   </xsl:template>
 
-  <xsl:template match="ezxhtml5:strong">
+  <xsl:template match="ezxhtml5:strong | ezxhtml5:b">
     <emphasis>
       <xsl:attribute name="role">strong</xsl:attribute>
       <xsl:if test="@class">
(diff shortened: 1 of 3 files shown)


src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -747,4 +747,22 @@
     <xsl:value-of select="translate( substring-before( substring-after( concat( substring-after( $style, $property ), ';' ), ':' ), ';' ), ' ', '' )"/>
   </xsl:template>
 
+  <!-- Some fallbacks to handle translating inline formatting in span tags for copy&paste and import use cases -->
+  <xsl:template match="ezxhtml5:span[not(@data-ezelement) and contains(@style,'bold')]">
```
alongosz: Not a huge fan of that.
POV ping @dew326: Can we rely on a CSS style containing "bold" phrase to always convert entire `<span>` as `<strong>` for OE format?

andrerom: I'll let @dew326 confirm. But afaik we don't use it, it would come from copy&paste or import cases, and in those cases there might be lots of styles at the same time, so in this case as the test shows the last one wins. I'm not aiming for perfection here just some more flexibility in Html5edit input parsing.

From what I could see `bold` and `bolder` are the only words in CSS terminology that will match this.

dew326: From the AE perspective, it doesn't matter. We are not using this and AE will never produce such format.

andrerom: @alongosz WDYT? :)

PR #94, opened 2019-12-20
EZP-30033: Handled additional para inside itemized/ordered list

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-30033](https://jira.ez.no/browse/EZP-30033)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 1.1
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Currently, content copied from external sources (Word) and containing additional `<p>` tags crashes publishing in BO. This is caused by the conversion adding hard `<para>` without checking if there is already one. According to [DocBook documentation](https://tdg.docbook.org/tdg/4.5/para.html) `<para>` cannot be child on another `<para>`. `<span>` tags inherited from copying are converted properly.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -348,9 +348,16 @@
         </xsl:attribute>
       </xsl:if>
       <xsl:call-template name="ezattribute"/>
-      <para>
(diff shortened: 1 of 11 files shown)


konradoboza:
Let's have also @alongosz's opinion as well. :)

konradoboza:
ping @lserwatka

PR #95, opened 2019-12-23
Ported fix for EZP-29958 and EZP-29984 from kernel

andrerom:
| Question           | Answer
| **JIRA issue**     | [EZP-29984](https://jira.ez.no/browse/EZP-29984) & [EZP-29958](https://jira.ez.no/browse/EZP-29958)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 1.1+
| **BC breaks**      | no
| **Tests pass**     | yes?
| **Doc needed**     | no?

Ports over fixes from:
- EZP-29984: Fix LIKE support for * as wildcard across engines, deprecate % usage
- https://github.com/ezsystems/ezpublish-kernel/pull/2517

Which also happens to align RichText with how XmlText had been fixed for:
- "Creating paragraph longer than 32766 bytes in RichText Editor will fail with Solr Error"
- https://jira.ez.no/browse/EZP-29958

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/FieldType/RichText/SearchField.php
+++ b/src/lib/eZ/FieldType/RichText/SearchField.php
@@ -78,7 +78,7 @@ class SearchField implements Indexable
      *
      * @return string
      */
-    public static function extractShortText(DOMDocument $document)
+    public static function extractShortText(DOMDocument $document): string
     {
         $result = null;
         // try to extract first paragraph/tag
--- a/src/lib/eZ/FieldType/RichText/SearchField.php
+++ b/src/lib/eZ/FieldType/RichText/SearchField.php
@@ -36,7 +36,7 @@ class SearchField implements Indexable
         return [
             new Search\Field(
                 'value',
-                $this->extractShortText($document),
+                self::extractShortText($document),
                 new Search\FieldType\StringField()
             ),
             new Search\Field(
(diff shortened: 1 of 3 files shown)


src/lib/eZ/FieldType/RichText/SearchField.php
```diff
@@ -72,13 +72,16 @@ private function extractText(DOMNode $node)
...
      * Extracts short text content of the given $document.
      *
+     * @internal Only for use by RichText FieldType itself.
+     *
      * @param \DOMDocument $document
      *
      * @return string
      */
-    private function extractShortText(DOMDocument $document)
+    public static function extractShortText(DOMDocument $document)
```
alongosz:
```suggestion
    public static function extractShortText(DOMDocument $document): string
```
TBH I think I overlooked it when reviewing ezsystems/ezpublish-kernel#2517. To have a clean implementation here, it should be extracted to a common Service. But not insisting, because it's a port and it's simple. Just sayin' ;)

andrerom:
> Just a small strict type request:

added

micszo:
Regarding wildcard support in ezauthor should the Email field also be covered or only the Name field?

micszo:
Also I have discovered that searching for values in ezemail and ezisbn fields (regardless of wildcards) doesn't work at all, even without Solr and this change. I need to dig in JIRA if we have something for that already.

andrerom:
> Regarding wildcard support in ezauthor should the Email field also be covered or only the Name field?

In context of what changed in EZP-29984, LIKE and sort by fixes to align Solr and LegacySE for default fields, [only name](https://github.com/ezsystems/ezpublish-kernel/pull/2517/files#diff-ba3f952eab27348b602eea068feee765R97-R111). email should be possible on solr but would involve searching on [custom field](https://github.com/ezsystems/ezpublish-kernel/blob/7.5/eZ/Publish/Core/FieldType/Author/SearchField.php#L51-L55) and not the default one, so not within scope of EZP-29984.

> discovered that searching for values in ezemail and ezisbn fields (regardless of wildcards) doesn't work

sounds like something to report separately with affected LTS versions

micszo:
Reported separately: https://jira.ez.no/browse/EZP-31287.

PR #100, opened 2020-01-08
EZP-30130: Added handling of soft breaks inside nested formatting within paragraphs

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-30130](https://jira.ez.no/browse/EZP-30130)
| **Bug/Improvement**| yes
| **New feature**    | no
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

As we allow having soft breaks within clean paragraphs and we flush inline tags, we can also avoid validation crashes by checking line breaks inside nested `<span>` tags coming from external editors like Word. Behavior is the same as with `<p>text[soft_break]more text</p>` formatting.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -109,6 +109,26 @@
     </para>
   </xsl:template>
 
+  <xsl:template match="ezxhtml5:span">
+    <xsl:choose>
+      <xsl:when test="descendant::ezxhtml5:br">
+        <xsl:for-each select="node()">
+          <xsl:choose>
+            <xsl:when test="local-name( current() ) = 'br'">
+              <xsl:text>&#xA;</xsl:text>
+            </xsl:when>
+            <xsl:otherwise>
+              <xsl:apply-templates select="current()"/>
+            </xsl:otherwise>
+          </xsl:choose>
+        </xsl:for-each>
+      </xsl:when>
+      <xsl:otherwise>
+        <xsl:apply-templates/>
+      </xsl:otherwise>
+    </xsl:choose>
+  </xsl:template>
+
   <xsl:template match="ezxhtml5:pre">
     <xsl:element name="programlisting">
       <xsl:if test="@id">
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -110,23 +110,9 @@
   </xsl:template>
 
   <xsl:template match="ezxhtml5:span">
-    <xsl:choose>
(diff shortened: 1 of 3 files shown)


micszo:
What about pasting test with new lines onto lists? :) Should that be also covered in scope of this issue?

konradoboza:
Since newlines within itemized/ordered lists are currently not supported (without taking pasting from external sources into account), I think we should handle it as another case. I am not sure if this is that useful though. 😅 WDYT @andrerom @alongosz?

andrerom:
@alongosz Can you maybe have a _deeper_ look here? 👼 

Besides the question on breaks in lists, I'm also noticing that there are a lot of break logic in here which differs, and which might be possible to cleanup. From the looks of it they should all be able to reuse `<xsl:template name="breakline">`.

But first, why is not the break logic in `<xsl:template match="ezxhtml5:p">` affecting the nested stuff here? @konradoboza I'm assuming the span is within the p?

micszo:
I don't want to beat a dead horse but the same question is valid for pasting into tables. 🙃

konradoboza:
The only cleanup I can see is changing `<br>` template call within `<span>`, added in https://github.com/ezsystems/ezplatform-richtext/pull/100/commits/08ab1cb22f21d795d42bc35f83fb5c93db44fba2. The same code for paragraph wraps tags also with `<literallayout>` that's why it doesn't reuse `breakline` template. 

When it comes to lists: if we paste content from external source we end up calling `<ezxhtml5:p>`, otherwise we just insert `<para>` tag, ref: https://github.com/ezsystems/ezplatform-richtext/pull/94/files#diff-fce53d3785fdab7f6dead4e8b03c31f7R355-R360. To fix that we would need to also call that template `ezxhtml5:p` otherwise breaklines will be ignored. The same thing applies to `<th>` and `<td>`.

konradoboza:
To sum things up: I think we should unblock this one and handle newlines in lists/tables/other formatting affected as a followup.

andrerom:
> code for paragraph wraps tags also with <literallayout> that's why it doesn't reuse breakline template.

So I'm unsure as I'm no xsl master, but it seems that this is done here which is why I suggested engineering should have a pass _(probably right after this)_:
https://github.com/ezsystems/ezplatform-richtext/blob/08ab1cb22f21d795d42bc35f83fb5c93db44fba2/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl#L726-L731

Anyway, that's more for a followup maybe.

> To sum things up: I think we should unblock this one and handle newlines in lists/tables/other formatting affected as a followup.

+10, @SylvainGuittard would it be possible to schedule some work on improving richtext parsing for copy/paste to handle more cases? It will most likely benefit the work to enable nesting in tables in RTE 🥇

micszo:
Follow-up ticket: https://jira.ez.no/browse/EZP-31286.

konradoboza:
@lserwatka this one can be merged.

PR #102, opened 2020-01-15
EZEE-2914: Custom inline style does not update textarea after click on

mateuszdebinski:
| Question           | Answer
| **JIRA issue**     | [EZEE-2914](https://jira.ez.no/browse/EZEE-2914)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | master target 3.0
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

AlloyEditor.ButtonStylesListItem on click does not update textarea element, because function _onClick is protected and we cannot override, so we need an override render function, for add a new function into onClick in order to update textarea

**TODO**:
- [ ] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslistitem.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslistitem.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+
+export default class EzBtnStylesListItem extends AlloyEditor.ButtonStylesListItem {
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonStylesListItem
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        const className = this.props.name === this.props.activeStyle ? 'ae-toolbar-element active' : 'ae-toolbar-element';
+
+        return (
+            <button
+                className={className}
+                dangerouslySetInnerHTML={{ __html: this._preview }}
+                onClick={() => {
+                    this._onClick();
+                    this.fireCustomUpdateEvent();
+                }}
+                tabIndex={this.props.tabIndex}
+            />
+        );
+    }
+
+    fireCustomUpdateEvent() {
+        const nativeEditor = this.props.editor.get('nativeEditor');
+
+        nativeEditor.fire('customUpdate');
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslistitem.js
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonStylesListItem
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
+        const className = this.props.name === this.props.activeStyle ? 'ae-toolbar-element active' : 'ae-toolbar-element';
```
GrabowskiM:
```suggestion
        let className = 'ae-toolbar-element';
        if (this.props.name === this.props.activeStyle) {
          className += ' active'
         } 
```
so you won't repeat basic class name

dew326: This code is copied from the AE, let's keep it as it is.

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslistitem.js
```diff
@@ -0,0 +1,37 @@
...
+
+    fireCustomUpdateEvent() {
+        const nativeEditor = this.props.editor.get('nativeEditor');
+
+        nativeEditor.fire('customUpdate');
+    }
+}
+
+AlloyEditor.ButtonStylesListItem = AlloyEditor.EzBtnStylesListItem = EzBtnStylesListItem;
+eZ.addConfig('ezAlloyEditor.ezBtnStylesListItem', EzBtnStylesListItem);
```
dew326: Richtext bundle doesn't know about ezplatform-admin-ui. Please update this like in the other buttons.

mateuszdebinski: corrected

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslistitem.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+import EzBtnLink from "./ez-btn-link";
```
dew326:
```suggestion
```

PR #106, opened 2020-01-23
EZP-30222: Handled tables copied from external sources

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-30222](https://jira.ez.no/browse/EZP-30222)
| **Bug/Improvement**| yes
| **New feature**    | no
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

This PR  introduces a slightly different way of handling table width:
- introducing `normalizeWidth` function which handles `px` and `pt` units as these are the main ones used in tables,
- rounding width values, e.g. `110.5pt`,
- taking `style="border: [thickness] [style] [color];"` property from CSS into account and applying `border="1px"` to make things consistent with the tables generated by RTE by default.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -397,15 +397,13 @@
           </xsl:call-template>
         </xsl:variable>
         <xsl:if test="$width != ''">
+          <xsl:variable name="normalizedWidth">
+            <xsl:call-template name="normalizeWidth">
+              <xsl:with-param name="width" select="$width"/>
+            </xsl:call-template>
+          </xsl:variable>
           <xsl:attribute name="width">
-            <xsl:choose>
-              <xsl:when test="substring( $width, string-length( $width ) - 1 ) = 'px'">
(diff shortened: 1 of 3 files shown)


konradoboza:
@alongosz that would be correct but we are actually "losing" some data here:
- value `style="border: 1px solid #000;"` is converted to the default `border="1"` and lost as we don't handle border styling,
- value for width in`pt`, as we convert and normalize it to `px` to not change prior behavior.

alongosz:
> @alongosz that would be correct but we are actually "losing" some data here:
> 
> * value `style="border: 1px solid #000;"` is converted to the default `border="1"` and lost as we don't handle border styling,

Ok, that makes sense. Still, can we see a test case for output (X)HTML?

> * value for width in`pt`, as we convert and normalize it to `px` to not change prior behavior.

We can indeed keep the normalization as-is right now.

konradoboza:
I added some more robust handling of not supported units (others than `px`, `pt`, `%`) to make the whole solution a bit more bulletproof. Also, simplified rounding units as `round` function does casting to number by default. @andrerom @alongosz please take another look.

PR #107, opened 2020-01-23
EZP-31272: Handled anchors generated with name attribute

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-31272](https://jira.ez.no/browse/EZP-31272)
| **Bug/Improvement**| yes
| **New feature**    | no
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Currently, anchors pasted from external sources often contain `name` attribute: `<a name="test">`. Since we only support `id` attribute I decided to add a simple mapping that prevents RTE validation from crushing.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -258,22 +258,13 @@
   <xsl:template name="link.anchor">
     <xsl:param name="attribute"/>
     <anchor>
-      <xsl:choose>
-        <xsl:when test="boolean(//ezxhtml5:a[@name = preceding::ezxhtml5:a/@name or ./ezxhtml5:a/@name = following::ezxhtml5:a/@name] ) = false()">
-          <xsl:attribute name="xml:id">
-            <xsl:value-of select="$attribute"/>
-          </xsl:attribute>
-        </xsl:when>
-        <xsl:otherwise>
-          <xsl:message terminate="yes">
-            Anchor names must be unique. Check duplicates and re-publish the content.
-          </xsl:message>
-        </xsl:otherwise>
-      </xsl:choose>
+      <xsl:attribute name="xml:id">
+        <xsl:value-of select="$attribute"/>
+      </xsl:attribute>
     </anchor>
   </xsl:template>
 
-  <xsl:template match="ezxhtml5:a">
+  <xsl:template match="ezxhtml5:a[not(@name=preceding::ezxhtml5:a/@name)]">
     <xsl:choose>
       <xsl:when test="@href">
         <xsl:call-template name="link.href"/>
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -256,9 +256,10 @@
   </xsl:template>
 
   <xsl:template name="link.anchor">
+    <xsl:param name="attribute"/>
     <anchor>
       <xsl:attribute name="xml:id">
-        <xsl:value-of select="@id"/>
+        <xsl:value-of select="$attribute"/>
       </xsl:attribute>
     </anchor>
   </xsl:template>
@@ -269,7 +270,14 @@
         <xsl:call-template name="link.href"/>
       </xsl:when>
       <xsl:when test="@id">
-        <xsl:call-template name="link.anchor"/>
+        <xsl:call-template name="link.anchor">
+          <xsl:with-param name="attribute" select="@id"/>
+        </xsl:call-template>
+      </xsl:when>
+      <xsl:when test="@name">
+        <xsl:call-template name="link.anchor">
+          <xsl:with-param name="attribute" select="@name"/>
+        </xsl:call-template>
       </xsl:when>
       <xsl:otherwise>
         <xsl:message terminate="yes">
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -258,9 +258,18 @@
   <xsl:template name="link.anchor">
     <xsl:param name="attribute"/>
     <anchor>
-      <xsl:attribute name="xml:id">
-        <xsl:value-of select="$attribute"/>
-      </xsl:attribute>
+      <xsl:choose>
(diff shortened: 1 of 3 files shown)


tests/lib/eZ/RichText/Converter/Xslt/_fixtures/xhtml5/edit/lossy/008-anchor.xhtml5.edit.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<section xmlns="http://ez.no/namespaces/ezpublish5/xhtml5/edit">
+    <p>
+        <a name="top"/>some anchor with a name attribute.
```
alongosz: I thought the point here was to allow and map both `xml:id` to `id` and `xml:name` to `name`. The tests however suggest that id is swapped with name, which TBH doesn't make sense.

konradoboza: The `name` attribute isn't used for anchors anywhere, that's why I decided to just map it to the existing one which is `id`.

konradoboza: Actually there is not, fair point. 😅 I will change that.

konradoboza: @alongosz I checked [doc reference](https://docbook.org/schemas/sdocbook/elements/anchor.html) and tested (validation fails) mapping to `name` attribute and it turns out that it stands against schema for anchors most likely. Should we aim to store it under some other attribute?

konradoboza: As `name` attribute can be mapped only to `id` I added checking if the mapped id is unique compared to the rest of the anchors in the current document. If it's not, there is validation error thrown with a more explanatory message. Added in: https://github.com/ezsystems/ezplatform-richtext/pull/107/commits/5dabd8250ede38fa7c79eee869093dc18bf09241.

konradoboza:
@adamwojs, @andrerom could you please take a look? 🤞

andrerom:
> @adamwojs, @andrerom could you please take a look? 🤞

looks good, but also looks like we should add more cases in test coverage now, at least one for when it is invalid.

konradoboza:
The only way to test duplicated `name` attributes I see is invoking intentional validation fail. Please see the CI result for https://github.com/ezsystems/ezplatform-richtext/pull/107/commits/be2e8692f54749b420bb72699b8c9d5dc6fc4f80: https://travis-ci.org/ezsystems/ezplatform-richtext/jobs/646323756#L658

konradoboza:
I reworked and simplified validation to not fail but strip anchors having duplicated `name` attributes. Therefore I apply dedicated rule to all `<a>` tags to take only unique (`name`-wise) ones into account. WDYT @andrerom, @adamwojs?

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -256,20 +256,28 @@
...
     <anchor>
       <xsl:attribute name="xml:id">
-        <xsl:value-of select="@id"/>
+        <xsl:value-of select="$attribute"/>
       </xsl:attribute>
     </anchor>
   </xsl:template>
 
-  <xsl:template match="ezxhtml5:a">
+  <xsl:template match="ezxhtml5:a[not(@name=preceding::ezxhtml5:a/@name)]">
```
andrerom: This won't affect links or anchors with id right?

konradoboza: Indeed, only the name duplicates will be checked.

konradoboza: Side comment: to be fully bulletproof we would need to also check if there are no duplicates when combining both `id` and `name` attributes produced by RTE and external sources mixed together.

Example:
`<a name="test"/><a name="test"/>` ➡️ `<anchor id="test/>test` ✅ 
`<a name="test"/><a id="test"/>` ➡️ `<anchor id="test"><anchor id="test">` :x:

Solution:
```
<xsl:template match="a[not(@id=preceding::a/@id) and 
                       not(@name=preceding::a/@name) and
                       not(@id=preceding::a/@name) and
                       not(@name=preceding::a/@id)]">
```

tests/lib/eZ/RichText/Converter/Xslt/_fixtures/xhtml5/edit/lossy/008-anchor.xhtml5.edit.xml
konradoboza: This could be fixed this way: https://github.com/ezsystems/ezplatform-richtext/pull/107#discussion_r375926810.

konradoboza:
Note to QA: please perform also sanities on all kinds of links (internal/external) in RTE.

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
konradoboza: After internal sync with reviewers, we decided to keep the current solution as is.

micszo:
First of all, this comment is not related to these changes - happens on v2.5.8. 🙂 
I noticed console error `TypeError: Cannot read property 'target' of undefined` occurs when adding more than one link to any of the lists and when adding a link to a phrase in a tabel cell. 
The workaround, to get the RT menu back, is to reload the page (e.g. by saving).

konradoboza:
@micszo indeed, this doesn't seem to be related to this PR.

micszo:
Also links added to `Formatted` style are not saved. But it also occurs without these changes.

konradoboza:
@lserwatka this one seems to be ready for merge.

lserwatka:
Could you merge it up?

konradoboza:
Done in https://github.com/ezsystems/ezplatform-richtext/commit/a7026d7d87a2efd01d84fda6f599b26f1eb9fde4.

micszo:
FYI, issues are reported and linked to JIRA ticket from description.

PR #115, opened 2020-02-21
EZP-31356: Fixed styles dropdown selection on Firefox and Safari

barw4:
| Question           | Answer
| **JIRA issue**     | [EZP-31356](https://jira.ez.no/browse/EZP-31356)
| **Bug**| yes
| **New feature**    | no
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     |no

The focus event which is hardcoded into the AlloyEditor and attached to the styles dropdown list is not working correctly on Firefox and Safari. The reasoning for the solution was to implement `ButtonDropdown` component without this `focus` event, which does more harm than good and inject it into newly created `EzButtonStylesList` component which overwrites the default AlloyEditor's `ButtonStylesList` in order to prevent other components of being mutated.

**TODO**:
- [x] Fix a bug.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
@@ -2,7 +2,6 @@ import React from 'react';
 import AlloyEditor from 'alloyeditor';
 
 export default class EzBtnDropdown extends AlloyEditor.ButtonDropdown {
-
     render() {
         return React.createElement("div", {
             className: "ae-dropdown ae-arrow-box ae-arrow-box-top-left",
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
@@ -2,12 +2,22 @@ import React from 'react';
 import AlloyEditor from 'alloyeditor';
 
 export default class EzBtnDropdown extends AlloyEditor.ButtonDropdown {
+    /**
+     * Lifecycle. Renders the UI of the button.
+     *
+     * @instance
+     * @memberof ButtonDropdown
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
     render() {
-        return React.createElement("div", {
-            className: "ae-dropdown ae-arrow-box ae-arrow-box-top-left",
-            onKeyDown: this.handleKey,
-            tabIndex: "0",
-        }, React.createElement("ul", {className: "ae-listbox", role: "listbox"}, this.props.children))
+        return (
+            <div className="ae-dropdown ae-arrow-box ae-arrow-box-top-left" onKeyDown={this.handleKey} tabIndex="0">
+                <ul className="ae-listbox" role="listbox">
+                    {this.props.children}
+                </ul>
+            </div>
+        );
     }
 }
 
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+
+export default class EzBtnDropdown extends AlloyEditor.ButtonDropdown {
+
+    render() {
+        return React.createElement("div", {
+            className: "ae-dropdown ae-arrow-box ae-arrow-box-top-left",
+            onKeyDown: this.handleKey,
+            tabIndex: "0",
+        }, React.createElement("ul", {className: "ae-listbox", role: "listbox"}, this.props.children))
+    }
+}
+
+const eZ = (window.eZ = window.eZ || {});
+
+eZ.ezAlloyEditor = eZ.ezAlloyEditor || {};
+eZ.ezAlloyEditor.EzBtnDropdown = EzBtnDropdown;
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-dropdown.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+
+export default class EzBtnDropdown extends AlloyEditor.ButtonDropdown {
+
```
GrabowskiM: ennecessary empty line

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import AlloyEditor from 'alloyeditor';
+import EzButtonDropdown from './ez-btn-dropdown';
+
+export default class EzButtonStylesList extends AlloyEditor.ButtonStylesList {
+
```
GrabowskiM: unnecessary empty line

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
```diff
@@ -0,0 +1,30 @@
...
+    render() {
+        var e;
```
GrabowskiM: let/const instead of var

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
```diff
@@ -0,0 +1,30 @@
...
+        return this.props.showRemoveStylesItem && (e = React.createElement(AlloyEditor.ButtonStylesListItemRemove, {
```
dew326: Can you use the JSX syntax?

barw4: @dew326

> To fix this in 2.5 it should be opened also to admin-ui bundle
Sure, I'm aware, after its fine here I'm gonna post 2nd PR to admin-ui-bundle

> Can you use the JSX syntax?
Yes, however, this `render` method is taken directly out of AlloyEditor, I wanted to keep it the same, should I reformat it?

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
barw4: @GrabowskiM 

this method is taken directly out of AlloyEditor as I mentioned here: https://github.com/ezsystems/ezplatform-richtext/pull/115#discussion_r382578977

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
dew326: From where did you copy this method?

barw4:
```
createReactClass({
                displayName: "ButtonStylesList"
```
Its basically this code fragment of alloy-editor-no-react-min.js - I know its minified, but its just render method of `ButtonStylesList` component.

dew326: If you want to copy something please always take the source:
https://github.com/liferay/alloy-editor/blob/1.x/src/ui/react/src/components/buttons/button-styles-list.jsx

src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-styleslist.js
```diff
@@ -0,0 +1,43 @@
...
+export default class EzButtonStylesList extends AlloyEditor.ButtonStylesList {
+    /**
+     * Lifecycle. Renders the UI of the list.
+     *
+     * @instance
+     * @memberof ButtonStylesList
+     * @method render
+     * @return {Object} The content which should be rendered.
+     */
+    render() {
```
dew326: I can't spot any meaning change in this file in comparison to the original. Why this file was overridden?

I think adding:
```js
AlloyEditor.ButtonDropdown = AlloyEditor.EzBtnDropdown = EzBtnDropdown;
```
In the `ez-btn-dropdown.js` should do the trick.

barw4: @dew326, that's what I did at the beginning, however, I'm not sure overriding whole `ButtonDropdown` is a good idea, because some other elements which work correctly may depend on it.

The single change is here at line 22: 
`            <EzButtonDropdown {...this.props}>
`

So the new `ButtonDropdown` is used **only** in `ButtonStylesList`, not everywhere else.

dew326: So if the other components work well with the focus why this one is not?

barw4: Because other components are not extendable like ButtonStylesList and they don't have long lists of items, which may cause problems. So basically, the only dropdown which may cause this issue is this one as there are many items on the list.

dew326:
Please create PR for 2.5

barw4:
> Please create PR for 2.5

@dew326  already done here: https://github.com/ezsystems/ezplatform-admin-ui/pull/1256

PR #119, opened 2020-03-13
EZP-30545: Provided way to define buttons in AlloyEditor toolbars

Nattfarinn:
| Question           | Answer
| **JIRA issue**     | [EZP-30545](https://jira.ez.no/browse/EZP-30545)
| **Bug/Improvement**| yes
| **New feature**    | yes
| **Target version** | `master`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | yes

Example configuration:

```yaml
system:
    admin_group:
        fieldtypes:
            ezrichtext:
                toolbars:
                    my_toolbar:
                        buttons:
                            my_last_button:
                                priority: -100
                            hidden_button:
                                visible: false
                            first_button:
                                priority: 50
                            button_with_defaults:
                                priority: 0
                                visible: true
```

**For QA**
All toolbars should be in the same place and with the same buttons by default
We need to test if it is possible to change the order of buttons in toolbars (by priority) and hide some buttons (by visible), also we need to test if it is possible to add new buttons into the toolbar by the yml file.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
+++ b/src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
@@ -30,6 +30,9 @@ class RichText extends AbstractFieldTypeParser
     private const ATTRIBUTE_TYPE_STRING = 'string';
     private const ATTRIBUTE_TYPE_NUMBER = 'number';
 
+    private const TOOLBARS_NODE_KEY = 'toolbars';
+    public const TOOLBARS_SA_SETTINGS_ID = 'fieldtypes.ezrichtext.' . self::TOOLBARS_NODE_KEY;
+
     // constants common for OE custom classes and data attributes configuration
     private const ELEMENT_NODE_KEY = 'element';
     private const DEFAULT_VALUE_NODE_KEY = 'default_value';
@@ -188,6 +191,32 @@ class RichText extends AbstractFieldTypeParser
                 ->scalarPrototype()->end()
             ->end();
 
+        // RichText Toolbars configuration (defines list of Toolbars and Buttons enabled for current SiteAccess scope)
+        $nodeBuilder
+            ->arrayNode(self::TOOLBARS_NODE_KEY)
+                ->useAttributeAsKey('name')
+                ->info('List of Toolbars and Buttons enabled for current SiteAccess scope.')
+                ->prototype('array')
+                    ->children()
+                        ->arrayNode('buttons')
+                            ->useAttributeAsKey('name')
+                            ->prototype('array')
+                                ->children()
+                                    ->booleanNode('visible')
+                                        ->info('Is button visible on toolbar?')
+                                        ->defaultTrue()
+                                    ->end()
+                                    ->integerNode('priority')
+                                        ->info('Defines order in which buttons appear (255 .. -255).')
+                                        ->defaultValue(0)
+                                    ->end()
+                                ->end()
+                            ->end()
+                        ->end()
+                    ->end()
+                ->end()
+            ->end();
+
         $this->buildOnlineEditorConfiguration($nodeBuilder);
     }
 
@@ -236,6 +265,7 @@ class RichText extends AbstractFieldTypeParser
             $onlineEditorSettingsMap = [
                 self::CLASSES_NODE_KEY => self::CLASSES_SA_SETTINGS_ID,
                 self::ATTRIBUTES_NODE_KEY => self::ATTRIBUTES_SA_SETTINGS_ID,
+                self::TOOLBARS_NODE_KEY => self::TOOLBARS_SA_SETTINGS_ID,
             ];
             foreach ($onlineEditorSettingsMap as $key => $settingsId) {
                 if (isset($scopeSettings['fieldtypes']['ezrichtext'][$key])) {
@@ -250,6 +280,7 @@ class RichText extends AbstractFieldTypeParser
     {
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.custom_tags', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.custom_styles', $config);
+        $contextualizer->mapConfigArray(self::TOOLBARS_SA_SETTINGS_ID, $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.output_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.edit_custom_xsl', $config);
         $contextualizer->mapConfigArray('fieldtypes.ezrichtext.input_custom_xsl', $config);
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-buttons.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-buttons.js
@@ -1,0 +1,53 @@
+import AlloyEditor from 'alloyeditor';
+
+export default class EzConfigButtonsBase {
+    getButtons(config) {
+        const toolbarConfig = window.eZ.richText.alloyEditor.toolbars[this.name];
+
+        if (!toolbarConfig) {
+            return [];
+        }
+
+        const buttons = [...toolbarConfig.buttons];
+        const attributesEditIndex = buttons.indexOf('ezattributesedit');
+        const stylesIndex = buttons.indexOf('ezstyles');
+
+        if (attributesEditIndex > -1) {
+            buttons[attributesEditIndex] = this.getEditAttributesButton(config);
+        }
+
+        if (stylesIndex > -1) {
+            buttons[stylesIndex] = this.getStyles(config.customStyles);
+        }
+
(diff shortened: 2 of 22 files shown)


src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
```diff
@@ -188,6 +191,29 @@ public function addFieldTypeSemanticConfig(NodeBuilder $nodeBuilder)
                 ->scalarPrototype()->end()
             ->end();
 
+        // RichText Toolbars configuration (defines list of Toolbars and Buttons enabled for current SiteAccess scope)
+        $nodeBuilder
+            ->arrayNode(self::TOOLBARS_NODE_KEY)
+                ->info('List of RichText Custom Tags enabled for the current scope. The Custom Tags must be defined in ezpublish.ezrichtext.custom_tags Node.')
```
alongosz: How is it related to Custom Tags?

Nattfarinn: It doesn't. Fixed.

tests/lib/Configuration/Provider/AlloyEditorTest.php
```diff
@@ -68,11 +69,165 @@ public function testGetConfiguration(): void
...
+    {
+        $configResolver = $this->createMock(ConfigResolverInterface::class);
+
+        $configResolver
+            ->method('hasParameter')
+            ->willReturn(true);
+
+        $configResolver
+            ->method('getParameter')
+            ->willReturnCallback(static function ($paramName) {
```
alongosz:
```suggestion
            ->willReturnCallback(static function (string $paramName): ?array {
```

tests/lib/Configuration/Provider/AlloyEditorTest.php
```diff
@@ -68,11 +69,165 @@ public function testGetConfiguration(): void
...
```
alongosz:
```suggestion
            ->willReturnCallback(static function (string $paramName): ?array {
```

tests/lib/Configuration/Provider/AlloyEditorTest.php
```diff
@@ -68,11 +69,165 @@ public function testGetConfiguration(): void
...
```
alongosz: AFAICS:
```suggestion
            ->willReturnCallback(static function (string $paramName): ?array {
```

tests/lib/Configuration/Provider/AlloyEditorTest.php
```diff
@@ -68,11 +69,165 @@ public function testGetConfiguration(): void
...
+                                ],
+                            ],
+                        ],
+                    ],
+                    RichText::CLASSES_SA_SETTINGS_ID => [],
+                    RichText::ATTRIBUTES_SA_SETTINGS_ID => [],
+                ];
+
+                return $map[$paramName] ?? null;
+            });
```
alongosz: This block is very similar in 3 places AFAICS. Could you extract it to common method with `buttons` part of config array as parameter, e.g. 
```php
private function buildTestToolbarReturnCallback(array $buttonsConfig): callable
```
?

Other approaches are of course welcomed as long as they avoid repetitions.

src/bundle/DependencyInjection/Configuration/Parser/FieldType/RichText.php
```diff
@@ -188,6 +191,31 @@ public function addFieldTypeSemanticConfig(NodeBuilder $nodeBuilder)
...
+                        ->arrayNode('buttons')
+                            ->useAttributeAsKey('name')
+                            ->prototype('array')
+                                ->children()
+                                    ->booleanNode('visible')
+                                        ->info('Is button visible on toolbar?')
+                                        ->defaultTrue()
+                                    ->end()
+                                    ->integerNode('priority')
+                                        ->info('Defines order in which buttons appear (255 .. -255).')
```
ViniTou: - 255 to 255?

Nattfarinn: Range can be wider, but -FF and +FF is a standard for Symfony

ViniTou: I understand range, I meant that `-255 to 255` is much more readable than `255 to -255` ;)

Nattfarinn: And yet, it's incorrect from priority point of view because suggests -255 will be first when in fact will be last. :)

ViniTou: least important to most important ;)

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-buttons.js
```diff
@@ -0,0 +1,53 @@
...
+export default class EzConfigButtonsBase {
+    getButtons(config) {
+        const toolbarConfig = window.eZ.richText.alloyEditor.toolbars[this.name];
+
+        if (!toolbarConfig) {
+            return [];
+        }
+
+        const buttons = [...toolbarConfig.buttons];
+        const attributesEditIndex = buttons.findIndex((button) => button === 'ezattributesedit');
```
GrabowskiM: could be indexOf instead of findIndex
```suggestion
        const attributesEditIndex = buttons.indexOf('ezattributesedit');
```

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-buttons.js
```diff
@@ -0,0 +1,53 @@
...
+        const stylesIndex = buttons.findIndex((button) => button === 'ezstyles');
```
GrabowskiM: same as above ^

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
```diff
@@ -1,21 +1,12 @@
...
-            'ezmoveup',
-            'ezmovedown',
-            editAttributesButton,
-            'tableHeading',
-            'ezembedinline',
-            'ezanchor',
-            'eztableremove',
-            ...config.extraButtons[this.name],
-        ];
+        this.name = this.getConfigName();
```
GrabowskiM: Are these two needed if you're calling `super(config)` already?

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/ez-custom-tag.js
```diff
@@ -4,24 +4,16 @@ export default class EzCustomTagConfig extends EzConfigBase {
...
-        ];
-        const customButtons = config.alloyEditor.toolbarButtons;
-        const buttons = customButtons && customButtons.length ? customButtons : defaultButtons;
-
         this.name = config.name;
-        this.buttons = buttons;
 
-        this.addExtraButtons(config.extraButtons);
+        const buttons = this.getButtons(config);
+        const customTagEditIndex = buttons.findIndex((button) => button === 'ezcustomtagedit');
```
GrabowskiM: could be indexOf instead of findIndex
```suggestion
        const customTagEditIndex = buttons.indexOf('ezcustomtagedit');
```

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/ez-inline-custom-tag.js
```diff
@@ -4,15 +4,16 @@ export default class EzInlineCustomTagConfig extends EzConfigBase {
...
-        const defaultButtons = [editButton, 'ezblockremove'];
-        const customButtons = config.alloyEditor.toolbarButtons;
-        const buttons = customButtons && customButtons.length ? customButtons : defaultButtons;
-
         this.name = config.name;
-        this.buttons = buttons;
 
-        this.addExtraButtons(config.extraButtons);
+        const buttons = this.getButtons(config);
+        const customTagEditIndex = buttons.findIndex((button) => button === 'ezcustomtagedit');
```
GrabowskiM: could be indexOf instead of findIndex
```suggestion
        const customTagEditIndex = buttons.indexOf('ezcustomtagedit');
```

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
dew326: But the parent doesn't have this assigned.

Nattfarinn:
@katarzynazawada Configuration part fixed: 

```yaml
            fieldtypes:
                ezrichtext:
                    toolbars:
                        paragraph:
                            buttons:
                                ezmoveup:
                                    priority: -100
                                ezmovedown:
                                    visible: false
```

Results in:

Zrzut ekranu 2020-03-23 o 17 46 29

PR #141, opened 2020-04-17
EZP-30861: Handled line breaks inside headings

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-30861](https://jira.ez.no/browse/EZP-30861)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `1.1`, `2.0`, `master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Currently, there is another inconsistent scenario when line breaks appear inside headings (`<h1> - <h6>`). As `literallayout` tag responsible for correctly displaying breaklines cannot be placed inside `<title>` tag (according to DocBook schema), I adjusted `docbook.rng` to reflect this case. As `db.all.inlines` rules are used in many places, I decided to add `literallayout` only to `title` element to avoid potential issues. XSL logic is basically about reusing existing rules for the paragraph - we don't want to add `literallayout` to `title` when there are no line breaks, so those scenarios are left untouched.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
+++ b/src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
@@ -1120,7 +1120,10 @@
         <a:documentation>The text of the title of a section of a document or of a formal block-level element</a:documentation>
         <ref name="db.title.attlist"/>
         <zeroOrMore>
-          <ref name="db.all.inlines"/>
+          <choice>
+            <ref name="db.all.inlines"/>
+            <ref name="db.literallayout"/>
+          </choice>
         </zeroOrMore>
       </element>
     </define>
--- a/src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
+++ b/src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
@@ -1120,10 +1120,7 @@
         <a:documentation>The text of the title of a section of a document or of a formal block-level element</a:documentation>
         <ref name="db.title.attlist"/>
         <zeroOrMore>
-          <choice>
-            <ref name="db.all.inlines"/>
-            <ref name="db.literallayout"/>
-          </choice>
+          <ref name="db.all.inlines"/>
         </zeroOrMore>
       </element>
     </define>
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -315,7 +315,25 @@
         </xsl:if>
       </xsl:if>
       <xsl:call-template name="ezattribute"/>
-      <xsl:apply-templates/>
+      <xsl:choose>
+        <xsl:when test="descendant::ezxhtml5:br">
+          <literallayout class="normal">
+            <xsl:for-each select="node()">
+              <xsl:choose>
+                <xsl:when test="local-name( current() ) = 'br'">
+                  <xsl:text>&#xA;</xsl:text>
+                </xsl:when>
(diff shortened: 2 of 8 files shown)


src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
```diff
@@ -1120,7 +1120,10 @@
         <a:documentation>The text of the title of a section of a document or of a formal block-level element</a:documentation>
         <ref name="db.title.attlist"/>
         <zeroOrMore>
-          <ref name="db.all.inlines"/>
+          <choice>
+            <ref name="db.all.inlines"/>
+            <ref name="db.literallayout"/>
+          </choice>
```
andrerom, line 1126: Asking this is thread so we can close it if irrelevant, but @alongosz do you remember if we have adaptions on this schema and are diverging from docbook spec on other things?

Either way, does:
- newer versions of docbook fix this? 
- or are there other ways we could deal with new lines in headers?

Another question: We might need to adapt ezplatform-xmltext-fieldtpye migratiomn for this or? (line breaks where supported in headers there too)

alongosz: > Another question: We might need to adapt ezplatform-xmltext-fieldtype migratiomn for this or? (line breaks where supported in headers there too)

Hmm, but did it work there then? Didn't we use there the same RelaxNG files? It should fail the same way. Or was XmlText spec different one? I'm sorry I don't remember this, hard to keep track of things, especially related to eZ Publish.

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -315,7 +315,25 @@
...
+                  <xsl:apply-templates select="current()"/>
+                </xsl:otherwise>
+              </xsl:choose>
+            </xsl:for-each>
+          </literallayout>
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:apply-templates/>
+        </xsl:otherwise>
+      </xsl:choose>
```
alongosz, line 336: wasn't there some "macro" a.k.a template for this already? Feels like it's a repetition.

src/lib/eZ/RichText/Resources/schemas/docbook/docbook.rng
andrerom: Unsure, maybe @vidarl remembers

konradoboza: According to the latest DocBook (5.2) documentation `literallayout  is still not mentioned among possible children for `title`, ref: https://tdg.docbook.org/tdg/5.2/title.html. I reverted customizing DocBook schema in `docbook.rng` and moved it to `ezpublish.rng` in 07fa7018c70bab79f11522b28e010cd12f845c43. As agreed, `literallayout` is an inline element, so I just override `db.all.inlines` section to take it into account.

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
konradoboza: Fixed in https://github.com/ezsystems/ezplatform-richtext/pull/141/commits/07fa7018c70bab79f11522b28e010cd12f845c43. After investigating potential cleanup scenarios and internal sync with @alongosz we agreed that adding another `breaklineWithLiterallayout` template is a way to go.

lserwatka:
Could you merge it up?

PR #148, opened 2020-05-15
EZP-31632: Floating table toolbar

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-31632](https://jira.ez.no/browse/EZP-31632)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Right now table toolbar has fixed positioning, and in some cases it brings inconvenience.

**Steps to reproduce:**
1. Open any content with RichtText field for editing
2. Insert a big table (50 rows) into RichText field
3. Scroll to the last table row and try to insert a new row after it

**Expected result**
The table toolbar should be shown on the screen, and new row should be inserted without any scrolling.

**Actual result**
The table toolbar remains at the start of the table. And the editor needs to scroll to it, to add a new row.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
@@ -10,9 +10,8 @@ export default class EzConfigFixedBase extends EzConfigBase {
         const editorRect = editor.element.getClientRect();
         const toolbarHeight = toolbar ? toolbar.getBoundingClientRect().height : 0;
         const shouldBeFixed = editorRect.top - toolbarHeight - 2 * TOOLBAR_OFFSET < 0;
-        const header = document.querySelector('.ez-edit-header__content-type-name');
         const top = shouldBeFixed
-            ? TOOLBAR_OFFSET + (header ? header.offsetHeight : 0)
+            ? TOOLBAR_OFFSET
             : editorRect.top + editor.element.getWindow().getScrollPosition().y - toolbarHeight - TOOLBAR_OFFSET;
 
         if (toolbar) {
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
@@ -10,8 +10,9 @@ export default class EzConfigFixedBase extends EzConfigBase {
         const editorRect = editor.element.getClientRect();
         const toolbarHeight = toolbar ? toolbar.getBoundingClientRect().height : 0;
         const shouldBeFixed = editorRect.top - toolbarHeight - 2 * TOOLBAR_OFFSET < 0;
+        const header = document.querySelector('.ez-edit-header__content-type-name');
         const top = shouldBeFixed
-            ? TOOLBAR_OFFSET
+            ? TOOLBAR_OFFSET + (header ? header.offsetHeight : 0)
             : editorRect.top + editor.element.getWindow().getScrollPosition().y - toolbarHeight - TOOLBAR_OFFSET;
 
         if (toolbar) {
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
@@ -1,18 +1,18 @@
-import AlloyEditor from 'alloyeditor';
-import EzConfigButtonsBase from './base-buttons';
+import EzConfigBase from './base';
 
-export default class EzConfigTableBase extends EzConfigButtonsBase {
+export default class EzConfigTableBase extends EzConfigBase {
     constructor(config) {
         super(config);
 
         this.name = this.getConfigName();
         this.buttons = this.getButtons(config);
-
-        this.getArrowBoxClasses = AlloyEditor.SelectionGetArrowBoxClasses.table;
-        this.setPosition = AlloyEditor.SelectionSetPosition.table;
     }
 
     getConfigName() {
+        return '';
+    }
+
+    getArrowBoxClasses() {
         return '';
     }
 }
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
@@ -1,6 +1,6 @@
-import EzConfigFixedBase from './base-fixed';
+import EzConfigBase from './base';
 
-export default class EzConfigTableBase extends EzConfigFixedBase {
+export default class EzConfigTableBase extends EzConfigBase {
     constructor(config) {
         super(config);
 
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
@@ -1,6 +1,6 @@
-import EzConfigBase from './base';
+import EzConfigFixedBase from './base-fixed';
 
-export default class EzConfigTableBase extends EzConfigBase {
+export default class EzConfigTableBase extends EzConfigFixedBase {
     constructor(config) {
         super(config);
 
--- a/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
@@ -13,6 +13,6 @@ export default class EzConfigTableBase extends EzConfigFixedBase {
     }
 
     getArrowBoxClasses() {
-        return '';
+        return 'ae-toolbar-floating';
     }
 }
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-linkedit.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-linkedit.js
@@ -133,14 +133,23 @@ export default class EzBtnLinkEdit extends Component {
      */
     showUI() {
         const nativeEditor = this.props.editor.get('nativeEditor');
-
-        nativeEditor.fire('editorInteraction', {
+        const eventOptions = {
             editor: this.props.editor,
             selectionData: {
                 element: this.state.element,
                 region: this.getRegion(),
             },
-        });
+        };
+        const path = nativeEditor.elementPath();
+
+        if (path) {
+            eventOptions.nativeEvent = {
+                editor: this.props.editor,
+                target: path.lastElement.$,
+            };
+        }
+
+        nativeEditor.fire('editorInteraction', eventOptions);
     }
 
     /**
--- a/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-linkedit.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/buttons/ez-btn-linkedit.js
@@ -133,14 +133,9 @@ export default class EzBtnLinkEdit extends Component {
      */
     showUI() {
         const nativeEditor = this.props.editor.get('nativeEditor');
-        const element = nativeEditor.elementPath().lastElement;
 
         nativeEditor.fire('editorInteraction', {
             editor: this.props.editor,
-            nativeEvent: {
-                editor: this.props.editor,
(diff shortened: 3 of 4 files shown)


dew326:
I think this bug is also in the 2.5 LTS. Please create this fix in ezplatform-admin-ui to branch 1.5.

SerheyDolgushev:
@dew326 Changed base branch to `master`, as personally I think it is not a bugfix, but an improvement.

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-table.js
```diff
@@ -1,18 +1,18 @@
-import AlloyEditor from 'alloyeditor';
-import EzConfigButtonsBase from './base-buttons';
+import EzConfigBase from './base';
 
-export default class EzConfigTableBase extends EzConfigButtonsBase {
+export default class EzConfigTableBase extends EzConfigBase {
```
dew326, line 3: I'm not sure but `EzConfigFixedBase` could be better because as I understand this fix will work for `td` and it's at the top of the `td`. When user will use elements path and will select `table` there will be the same problem. With `EzConfigFixedBase` it will be fixed at the top o the screen if it should be outside.

Also, the second question is do we want to cover the table with toolbar (I know there was some clients that don't like when toolbar is covering what they wrote in previous line), but this is a question to @SylvainGuittard.

SylvainGuittard: I just tried the steps to reproduce and I understand the pain point here. 
I think it makes sense for the table toolbar to be visible by the user.

inakijv: Hi guys,
Adding the UX perspective here. I would go for a similar solution that the RTE Paragraph toolbar has right now, given this behavior was added after being user tested, I agree @dew326.
Maybe a potential option to make this work better would be also to remove the _fixed_ position from `ez-edit-header__content-type-name`.

SerheyDolgushev:
> Hello guys, any release date planned for this fix ?

@neiluJ meantime you can use https://packagist.org/packages/contextualcode/ezplatform-alloyeditor-table-tools

dew326:
For this to be merged we need:
1. Rebase with current master
2. Change the `EzConfigBase` to `EzConfigFixedBase` as this behavior is better. (https://github.com/ezsystems/ezplatform-richtext/pull/148#discussion_r426620997)

SerheyDolgushev:
> For this to be merged we need:
> 
> 1. Rebase with current master
> 2. Change the `EzConfigBase` to `EzConfigFixedBase` as this behavior is better. ([#148 (comment)](https://github.com/ezsystems/ezplatform-richtext/pull/148#discussion_r426620997))

1. Latest master is merged
2. Done: https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/402bc7ae27f7b8ed44be1d0bce8f272f8470a830

dew326:
Could you also explain this commit? https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/cff502a119e4cfc3efed12b9592a36ab1d394243
Seems like it fixing some bug, which is not only in master?

SerheyDolgushev:
@dew326 
Sorry, I`m not sure why https://github.com/ezsystems/ezplatform-richtext/commit/cff502a119e4cfc3efed12b9592a36ab1d394243 is a part of this PR, as it is related to https://github.com/ezsystems/ezplatform-richtext/pull/149. Most likely I realized it after https://github.com/ezsystems/ezplatform-richtext/pull/149 was merged, but still I do not remember why this commit was pushed to this PR. Also, I just pushed https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/b6760838ec6582c3190ba3fc5ca622343179949d. Which is directly related to https://github.com/ezsystems/ezplatform-richtext/pull/163.

Should both of those commits be removed from this PR? And should we create a separate PR for each of them?

Also please note https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/43f5ad800e00c616a561735e47efa4c6173ca2de. And the reason for this, is that when `EzConfigFixedBase` is used the toolbar is shown at the top of the table. But when `EzConfigBase` is used the toolbar is shown at the active table row. And that is the main purpose of this PR.

SerheyDolgushev:
Sorry, but right now I have a feeling like https://github.com/ezsystems/ezplatform-richtext/commit/cff502a119e4cfc3efed12b9592a36ab1d394243 and https://github.com/ezsystems/ezplatform-richtext/commit/b6760838ec6582c3190ba3fc5ca622343179949d are making all the things much more complicated and they should be removed from this PR. A separate PR will be submitted for them. I'm not going to remove them from this PR, but will just revert, to make sure their links will remain in the PR conversation messages.

SerheyDolgushev:
https://github.com/ezsystems/ezplatform-richtext/pull/165 was submitted, and it is absolutely unrelated to this PR.

dew326:
> Also please note [43f5ad8](https://github.com/ezsystems/ezplatform-richtext/commit/43f5ad800e00c616a561735e47efa4c6173ca2de). And the reason for this, is that when `EzConfigFixedBase` is used the toolbar is shown at the top of the table. But when `EzConfigBase` is used the toolbar is shown at the active table row. And that is the main purpose of this PR.

We had complaints about that the toolbar covers the previous text and this was the reason why we introduced this fixed toolbar. Yes, it will go to the top of a table but when the top of the table is outside of the screen it will be fixed to the top of the window.

SerheyDolgushev:
@dew326 Sorry, but I just did another test using `EzConfigFixedBase` using the case from this PR description. And the toolbar wasn't shown at all when "the top of the table is outside of the screen":

table-floating-toolbar

In that case, maybe something needs to be fixed in `EzConfigFixedBase`?

dew326:
It works for me:
Zrzut ekranu 2020-10-1 o 07 36 53

You have some fixed bar at the top, which I don't recall if we have something like this in our product. Maybe this bar is covering the alloyEditor toolbar?

SerheyDolgushev:
@dew326 `EzConfigFixedBase` was not working for me because of missing `ae-toolbar-floating` CSS class.

So `EzConfigFixedBase` was enabled: https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/c1fd697e18c5b9b6b207b7ca65020eba3cfe5609. And missing `ae-toolbar-floating` CSS was fixed: https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/2e6acfe1ae207581c0105a33ebf524a54ae18592.

Also I was checking this PR on 3.1.3 and found out that the toolbar is covered by "edit header", which seems to be not a case for latest master because of the new UI changes. It is fixed in https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/16d73b0774d198aba783143f447d83add4d0f447.

table-toolbar

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
```diff
@@ -10,8 +10,9 @@ export default class EzConfigFixedBase extends EzConfigBase {
         const editorRect = editor.element.getClientRect();
         const toolbarHeight = toolbar ? toolbar.getBoundingClientRect().height : 0;
         const shouldBeFixed = editorRect.top - toolbarHeight - 2 * TOOLBAR_OFFSET < 0;
+        const header = document.querySelector('.ez-edit-header__content-type-name');
```
dew326, line 13: I think we don't have this header in master - 3.2 so I think this is not needed.

dew326 (changes requested):

src/bundle/Resources/public/js/OnlineEditor/toolbars/config/base-fixed.js
SerheyDolgushev: Done in https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/914751514f163c8b0cb63f92828bd9e8e955cf9e

dew326 (approved):

lucasOsti (approved):

GrabowskiM (approved):

SylvainGuittard:
Thanks @SerheyDolgushev for your contribution. 

> I think this bug is also in the 2.5 LTS. Please create this fix in ezplatform-admin-ui to branch 1.5 or if you feel it more like improvement please change the base branch to master here because it's questionable if it is bug or improvement.

Sorry, I am late in the game. I agree with Darek. I consider this as a bug. IMHO, this fix should also go to 2.5LTS
Can we merge and backport or should we rebase again 1.5?

dew326:
In 2.5LTS we had AlloyEditor in the admin-ui bundle so another PR should be opened in this repo.
This PR should be rebased with the latest supported branch so with 2.1.

SerheyDolgushev:
@dew326 sorry, but if this is going to be based on `2.1` we need to put back https://github.com/ezsystems/ezplatform-richtext/pull/148/commits/16d73b0774d198aba783143f447d83add4d0f447. Previously I was asked to disabled that commit. Can you please confirm rebase on 2.1 and putting that commit back?

dew326:
Yes, if we want to merge this starting from the 2.5LTS we need to merge it to:
2.5 - PR in admin-ui
3.1 - as this is the latest support version (this PR), and later we will merge up this into the master.

In this way, we will have fix in 2.5LTS, 3.1 and 3.2.

SerheyDolgushev:
@dew326 this PR is based on `2.1` now, and the "Edit header" offset was enabled. Please let me know if any changes are required here.

dew326:
No, this PR is good.

dew326:
I will pass this PR to the QA.

@SerheyDolgushev Could you prepare the same fix for 2.5 (in admin-ui bundle)?

dew326:
Note for QA:
This PR should also fix customer request in 3.1 https://jira.ez.no/browse/EZP-32006

katarzynazawada (changes requested):
Toolbar is shown when scrolling but it skips for a while when selecting another cell in the table. See please: http://g.recordit.co/CxmDj9xf75.gif

SerheyDolgushev:
Sorry, but it seems like the lag is caused by the huge size of the test table. And it seems like that lag is not related to this PR. @dew326 have you had any chance to look into this? Maybe there should be a separate PR/issue for toolbar rendering performance on huge tables like the one from the last QA?

micszo (changes requested):
Hi @SerheyDolgushev 
I tested these changes on v3.1.4.

Even with a small table (5x5) the issue exists.

**Actual:** Toolbar is first aligned to the table cell then, after a while, toolbar is aligned to the table (to the left)
**Expected:**  Toolbar is aligned to the table (to the left, as for a long portion of text) from the beginning.

Regarding the bug https://issues.ibexa.co/browse/EZP-32006 we might need to deal with it separately. \cc @barw4

SerheyDolgushev:
Hello @micszo

Could you please test it now?

micszo:
Hi @SerheyDolgushev 

the sticky toolbar works fine now for long portions of text and large tables.

However the pop-up with settings for custom tags (youtube, facebook, twitter) is broken now.
Please see the recording: https://recordit.co/7cLmNWL37l (the complete animation didn't get recorded)

**Steps:**
1. Add e.g. youtube custom tag.
2. Fill URL.
3. Save
4. Try to open custom tag settings with edit pencil.

*Actual result:* The pop-up jumps out of the visible area.
*Expected result:* The pop-up is displayed above the custom tag.

SerheyDolgushev:
Hello @micszo can you please check all the possible usecases one more time? Also please note, that there are separate issues with the custom tag toolbar, which was fixed in https://github.com/ezsystems/ezplatform-richtext/pull/169.

micszo (approved):
The changes request in https://issues.ibexa.co/browse/EZP-31632 and https://issues.ibexa.co/browse/EZP-32006 are in place now.

Didn't find new issues with the toolbars this round.

Checked both block and inline custom tags.

Verified on eZ Platform EE v3.1.4 with diff. Also applied the diff for https://issues.ibexa.co/browse/EZP-32187.

lserwatka:
@dew326 could you merge it up?

PR #160, opened 2020-08-07
EZP-31774: redesign richtext

lucasOsti:
jira ref: https://jira.ez.no/browse/EZP-31774

### DIFF
--- a/src/bundle/Resources/public/scss/_custom-styles.scss
+++ b/src/bundle/Resources/public/scss/_custom-styles.scss
@@ -4,19 +4,34 @@
     div[class*=' ezstyle-'],
     div[data-ezelement='eztemplate'][data-eztype='style'] {
         margin-top: 0;
-        margin-bottom: 1rem;
-        padding: 0.5rem;
-        border-width: 0.5rem;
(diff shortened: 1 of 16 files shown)


src/bundle/Resources/public/scss/variables/colors.scss
```diff
@@ -1,3 +1,29 @@
+// Ibexa colors
+$ibexa-white: #FFF;
```
dew326, line 2: Please use `!default`

src/bundle/Resources/public/scss/_custom-styles.scss
```diff
@@ -4,19 +4,34 @@
...
     span[data-ezelement='eztemplateinline'][data-eztype='style'] {
-        background-color: $ez-ae-color-ground;
-        padding: 0 0.25rem;
+        background-color: $ibexa-color-base-pale;
+        padding: 0 calculateRem(4px);
         font-family: courier, monospace;
     }
 }
+
+.cke_widget_focused > .ez-custom-tag{
```
GrabowskiM, line 24: missing whitespace

PR #162, opened 2020-09-14
EZP-31868: improve richtext embed access check on 404 page

ITernovtsii:
| Question           | Answer
| **JIRA issue**     | [EZP-31868](https://jira.ez.no/browse/EZP-31868)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 1.1
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

1) Render richtext field in pagelayout (for example, in footer).
2) Put embed image in the richtext field.
3) Open valid URL to verify anonymous has access to the image.
3) Open some 404 URL

AR:
500 error returned after max nested level reached

ER:
404 Page properly displayed

Developer notes:
It's because there is no firewall for 404 page, and there is no token in token storage service.
So, it threw "AuthenticationCredentialsNotFoundException".
Subrequest tries to render an error page, which extends pagelayout, which tries to render the same richtext field, and it leads to an infinitive loop.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/eZ/RichText/Renderer.php
+++ b/src/bundle/eZ/RichText/Renderer.php
@@ -8,21 +8,16 @@ declare(strict_types=1);
 
 namespace EzSystems\EzPlatformRichTextBundle\eZ\RichText;
 
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Repository;
-use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use EzSystems\EzPlatformRichText\eZ\RichText\RendererInterface;
-use Psr\Log\NullLogger;
-use Symfony\Component\Templating\EngineInterface;
-use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
-use eZ\Publish\Core\MVC\Symfony\Security\Authorization\Attribute as AuthorizationAttribute;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
-use eZ\Publish\API\Repository\Exceptions\NotFoundException;
-use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
-use Symfony\Component\Security\Core\Exception\AccessDeniedException;
-use Symfony\Component\Security\Core\Exception\AuthenticationCredentialsNotFoundException;
-use Exception;
 use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+use Symfony\Component\Templating\EngineInterface;
 
 /**
  * Symfony implementation of RichText field type embed renderer.
@@ -38,6 +33,8 @@ class Renderer implements RendererInterface
     protected $repository;
 
     /**
+     * @deprecated since 2.1.2
+     *
      * @var \Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface
      */
     private $authorizationChecker;
@@ -83,16 +80,9 @@ class Renderer implements RendererInterface
     private $customStylesConfiguration;
 
     /**
-     * @param \eZ\Publish\API\Repository\Repository $repository
-     * @param \Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface $authorizationChecker
-     * @param \eZ\Publish\Core\MVC\ConfigResolverInterface $configResolver
-     * @param \Symfony\Component\Templating\EngineInterface $templateEngine
      * @param string $tagConfigurationNamespace
      * @param string $styleConfigurationNamespace
      * @param string $embedConfigurationNamespace
-     * @param \Psr\Log\LoggerInterface|null $logger
-     * @param array $customTagsConfiguration
-     * @param array $customStylesConfiguration
      */
     public function __construct(
         Repository $repository,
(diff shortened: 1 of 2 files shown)


src/bundle/eZ/RichText/Renderer.php
```diff
@@ -162,6 +164,24 @@ function (Repository $repository) use ($contentId) {
             );
 
             $isDenied = true;
+        } catch (AuthenticationCredentialsNotFoundException $e) {
```
kaff, line 167: it looks like this try/catch construction has grown a lot, it could be good to split it a bit into private functions

src/bundle/eZ/RichText/Renderer.php
```diff
@@ -216,6 +236,23 @@ public function renderLocationEmbed($locationId, $viewType, array $parameters, $
 
                 return null;
             }
+        } catch (AuthenticationCredentialsNotFoundException $e) {
+            // There is no firewall for 404 page, and we're unable to use "isGranted".
```
kaff, line 240: it looks like this try/catch construction has grown a lot, it could be good to split it a bit into private functions

kaff (commented):
It looks like you fixed some important issue. I'm wondering that are you sure we don't need any updates in tests or write a new one for this case?

ITernovtsiy:
@kaff Initially, I did it with minimal changes to make sure it's something Ibexa team consider to fix. Please review it now.
Changes:
- I refactored render methods to use a single `renderEmbed` function, to not duplicate code.
- used `PermissionResolver::canUser` instead of `AuthorizationChecker::isGranted`
- updated tests to test new methods
- deperecate `$authorizationChecker` argument and `checkLocation` function

Let me know if it's acceptable to do a small BC break here, and I'll remove `checkLocation`, `checkContent`(deprecated since 6.7) functions and `$authorizationChecker` dependency.
And I'll use dependency injectiion for `PermissionResolver` instead of loading it from `$this->repository->getPermissionResolver()`

Thanks.

alongosz (changes requested):
> Let me know if it's acceptable to do a small BC break here, and I'll remove `checkLocation`, `checkContent`(deprecated since 6.7) functions and `$authorizationChecker` dependency.

No, it's not acceptable, because we've released this code in eZ Platform 2.x. See [SemVer summary](https://semver.org/#summary) for details about why. We sometimes make exceptions, usually due to unsolvable cases otherwise. I see no reason to do it here. See that `checkContent` is gone in 2.x branches (eZ Platform 3.0). It has been removed in the SemVer-compliant fashion. Moreover I don't see `checkLocation` being deprecated in any version.

Diff is quite complicated and TBH I need to spend some time to understand what's happening here.

What I'm not following right now already is that you've reported this bug for 3.x, yet you're fixing it in RichText version corresponding to 2.5. Which is it then?

ITernovtsiy:
@alongosz 
> No, it's not acceptable

Good, as I expected) so we can leave it as is and deprecate in the next released version.

> Diff is quite complicated

Yes, diff is big, while real changes are small - using `PermissionResolver` instead of `AuthorizationChecker`.
A big part of changes is refactoring of exception catches, so code is not duplicated in two functions.

> What I'm not following right now already is that you've reported this bug for 3.x, yet you're fixing it in RichText version corresponding to 2.5. Which is it then?

Initially, I reproduced it in v3.1 and v3.0.
After I started this PR, I found that the same code used in v2.5, so I rebased it to target `1.1` branch.
Let me know if I should rebase this back to `v3.x` only.  _(personally, I need this for 3.1+)_

alongosz (changes requested):
@ITernovtsiy I'm having trouble reproducing this issue and I need to do so to understand what's happening here.
Could you give an example for
> Render richtext field in pagelayout (for example, in footer).

?

ITernovtsiy:
@alongosz Sure, with clean install, it can be reproduced in this way:
- in `pagelayout.html.twig`, add:
`{{ render(controller('App\\Controller\\TestController::footerInfo')) }}`
- Create a minimal controller (`src/Controller/TestController.php`) to load content and render template:
```
namespace App\Controller;

use eZ\Bundle\EzPublishCoreBundle\Controller;
use eZ\Publish\API\Repository\ContentService;
use Symfony\Component\HttpFoundation\Response;

class TestController extends Controller
{
    private ContentService $contentService;

    public function __construct(ContentService $contentService)
    {
        $this->contentService = $contentService;
    }

    public function footerInfo(): Response
    {
        return $this->render('@ezdesign/footer_info.html.twig', [
            'content' => $this->contentService->loadContent(1)
        ]);
    }
}
```
- Create template for richtext field rendering
`templates/themes/standard/footer_info.html.twig`
```
RichText field: {{ ez_render_field(content, "description") }}
```
- Create error templates, content can be the same for both:
`templates/bundles/TwigBundle/Exception/error500.html.twig`
`templates/bundles/TwigBundle/Exception/error404.html.twig`
```
{% extends "@ezdesign/pagelayout.html.twig" %}
{% block content %}Error 404 or 500{% endblock %}
```
- Embed image in `description` field of content with id=2
- Open some URL which leads to 404
- You will see max nested level error, instead of 404 page.

_I wrote instruction using existing project, so let me know if it's not reproducible with provided steps and I'll work on clean install locally and making sure it's reproducible_
Thanks.

alongosz:
@ITernovtsiy unfortunately I cannot reproduce the issue on 2.5. I tried to adapt the code for 2.5 meta: https://github.com/ezsystems/ezplatform/compare/2.5...alongosz:review-2.5/ezp-31868-test-case
For me embedded image on an error page (accessed via Symfony `_error` dev testing route) is displayed w/o any errors.

What I don't follow in your code is the fact that you use `content` block in error templates, yet pagelayout itself does not define that block. Moreover the flow here is a bit risky: you're extending pagelayout for the error pages which via subrequest performs heavy operation of rendering a controller, which in turn might result in any error, not just the one you're experiencing. No wonder that you get infinite loop for an error. While indeed the error should not occur in the first place, the error page should rather be ready to handle all errors.

alongosz (changes requested):
:point_up: Still unclear what's going on here.

ITernovtsiy:
@alongosz 
I installed clean v2.5 and I can reproduce this after small changes in your code:
- rename `app/Resources/TwigBundle/views/Exception/error400.html.twig` to `app/Resources/TwigBundle/views/Exception/error404.html.twig`
- make sure symfony is in prod mode
- open some random page, like `/page-not-found-404`

I understand that extending pagelayout for error is not ideal, but it's what we have as a client requirement.
Anyway, using pagelayout for 404 pages is quite common, for example - https://www.ibexa.co/page-not-found
So, it would be nice to be able to render richtext on 404 page.

PR #166, opened 2020-10-01
EZP-31231: Fixed storing align and anchor attributes for custom styles

barw4:
| Question           | Answer
| **JIRA issue**     | [EZP-31231](https://jira.ez.no/browse/EZP-31231)
| **Bug**| yes
| **New feature**    | no
| **Target version** | 1.1
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Align and anchor attributes were not storing and displaying properly when working with custom style, also additional option to render `id` attribute has been added in form of:

```
<h2 id="{{ id }}" style="text-align:{{ align }}; color:red">
    {{ content }}
</h2>
```

where the code above is some [custom style](https://doc.ezplatform.com/en/2.5/guide/extending/extending_online_editor/#custom-styles).

### QA
// maintainer update:
- Custom Tags and Styles share markup and codebase, so regressions should be checked for both of them.

**TODO**:
- [x] Fix a bug.
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -123,6 +123,7 @@ class Template extends Render implements Converter
      *     attribute => param,
      *     ...
      * ].
+     *
      * @return string[]
      */
     protected function getAttributesMap(): array
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -123,6 +123,7 @@ class Template extends Render implements Converter
      *     attribute => param,
      *     ...
      * ].
+     * @return string[]
      */
     protected function getAttributesMap(): array
     {
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -97,8 +97,12 @@ class Template extends Render implements Converter
         }
         $parameters['content'] = !empty($innerContent) ? $innerContent : null;
 
-        if ($template->hasAttribute('ezxhtml:align')) {
-            $parameters['align'] = $template->getAttribute('ezxhtml:align');
+        if ($template->hasAttribute('ezxhtml:textalign')) {
+            $parameters['align'] = $template->getAttribute('ezxhtml:textalign');
+        }
+
+        if ($template->hasAttribute('xml:id')) {
+            $parameters['id'] = $template->getAttribute('xml:id');
         }
 
         $content = $this->renderer->renderTemplate(
--- a/src/lib/eZ/RichText/Converter/Render/Template.php
+++ b/src/lib/eZ/RichText/Converter/Render/Template.php
@@ -97,12 +97,10 @@ class Template extends Render implements Converter
         }
         $parameters['content'] = !empty($innerContent) ? $innerContent : null;
 
-        if ($template->hasAttribute('ezxhtml:textalign')) {
-            $parameters['align'] = $template->getAttribute('ezxhtml:textalign');
-        }
-
-        if ($template->hasAttribute('xml:id')) {
-            $parameters['id'] = $template->getAttribute('xml:id');
+        foreach ($this->getAttributesMap() as $attribute => $param) {
+            if ($template->hasAttribute($attribute)) {
+                $parameters[$param] = $template->getAttribute($attribute);
+            }
         }
 
         $content = $this->renderer->renderTemplate(
@@ -117,6 +115,21 @@ class Template extends Render implements Converter
             $payload->appendChild($document->createCDATASection($content));
             $template->appendChild($payload);
         }
+    }
+
+    /**
+     * Provides attributes map in format:
+     * [
+     *     attribute => param,
+     *     ...
+     * ].
+     */
+    protected function getAttributesMap(): array
+    {
+        return [
+            'ezxhtml:align' => 'align',
+            'xml:id' => 'id',
+        ];
     }
 
     /**
--- a/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <xsl:stylesheet
-    xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
-    xmlns:docbook="http://docbook.org/ns/docbook"
-    xmlns:xlink="http://www.w3.org/1999/xlink"
-    xmlns:ezxhtml="http://ez.no/xmlns/ezpublish/docbook/xhtml"
-    xmlns:ezcustom="http://ez.no/xmlns/ezpublish/docbook/custom"
-    exclude-result-prefixes="docbook xlink ezxhtml ezcustom"
-    version="1.0">
+        xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+        xmlns:docbook="http://docbook.org/ns/docbook"
+        xmlns:xlink="http://www.w3.org/1999/xlink"
+        xmlns:ezxhtml="http://ez.no/xmlns/ezpublish/docbook/xhtml"
+        xmlns:ezcustom="http://ez.no/xmlns/ezpublish/docbook/custom"
+        exclude-result-prefixes="docbook xlink ezxhtml ezcustom"
+        version="1.0">
   <xsl:output indent="yes" encoding="UTF-8"/>
   <xsl:variable name="outputNamespace" select="''"/>
 
@@ -688,9 +688,14 @@
         <xsl:value-of select="@ezxhtml:class"/>
       </xsl:attribute>
     </xsl:if>
-    <xsl:if test="@ezxhtml:align">
-      <xsl:attribute name="data-ezalign">
-        <xsl:value-of select="@ezxhtml:align"/>
+    <xsl:if test="@xml:id">
+      <xsl:attribute name="id">
+        <xsl:value-of select="@xml:id"/>
+      </xsl:attribute>
+    </xsl:if>
+    <xsl:if test="@ezxhtml:textalign">
+      <xsl:attribute name="style">
+        <xsl:value-of select="concat( 'text-align:', @ezxhtml:textalign, ';' )"/>
       </xsl:attribute>
     </xsl:if>
   </xsl:template>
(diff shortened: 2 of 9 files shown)


src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -722,11 +722,24 @@
...
+      <xsl:if test="@id">
+        <xsl:attribute name="xml:id">
+          <xsl:value-of select="@id"/>
         </xsl:attribute>
       </xsl:if>
+      <xsl:if test="contains( @style, 'text-align:' )">
+        <xsl:variable name="textAlign">
+          <xsl:call-template name="extractStyleValue">
+            <xsl:with-param name="style" select="@style"/>
+            <xsl:with-param name="property" select="'align'"/>
```
konradoboza, line 734: Why not narrow this down to the specific rule:
```suggestion
            <xsl:with-param name="property" select="'text-align'"/>
```
?

konradoboza (approved):
LGTM 👍

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
barw4: Yeah, that's a good idea.

webhdx (approved):

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
```diff
@@ -688,9 +688,14 @@
...
+    <xsl:if test="@xml:id">
+      <xsl:attribute name="id">
+        <xsl:value-of select="@xml:id"/>
+      </xsl:attribute>
+    </xsl:if>
     <xsl:if test="@ezxhtml:align">
-      <xsl:attribute name="data-ezalign">
-        <xsl:value-of select="@ezxhtml:align"/>
+      <xsl:attribute name="style">
+        <xsl:value-of select="concat( 'text-align:', @ezxhtml:align, ';' )"/>
```
alongosz, line 698: For edit mode it still should be data-ezalign AFAIK. @dew326?

src/lib/eZ/RichText/Converter/Render/Template.php
```diff
@@ -115,6 +117,21 @@ protected function processTemplate(DOMDocument $document, DOMXPath $xpath, DOMEl
         }
     }
 
+    /**
+     * Provides attributes map in format:
+     * [
+     *     attribute => param,
+     *     ...
+     * ].
```
webhdx, line 125: We tend to add `@return` attribute to all functions returning `array` type. So here you could add `@return string[]`

Nattfarinn, line 125: In case of `array` return or argument for scalar (or any non-nested structure) values, it is good practice to provide a docblock type notation like: `@return string[]`.

webhdx (approved):

Nattfarinn (approved):

alongosz (changes requested):
Not sure why you needed to change edit mode output, we only needed to expose that for view output custom template.
The change to edit output is a BC break.

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
konradoboza: Apart from front-end remark, out of curiosity: why changing edit mode output is considered BC break if those attributes weren't stored in the database in the first place?

dew326: Yes, in alloyEditor we have `data-ezalign`.

barw4: @alongosz it breaks edit otherwise, that's why I've changed it. Unchanged result:
image

dew326: That's because in our edit mode we have defined float to ezalign right. https://github.com/ezsystems/ezplatform-richtext/blob/master/src/bundle/Resources/public/scss/_alloyeditor-ez.scss#L1280

Why it is not "broken" when you set this property in the AlloyEditor?

barw4: @dew326 that's how headings, paragraphs and pre works. They don't utilize {{data-ezalign}}, they operate only on {{style="text-align"}}, they would break as well when using {{data-ezalign}}.

alongosz: > @alongosz apart from the front-end remark, out of curiosity: why changing edit mode output is considered a BC break if those attributes weren't stored in the database in the first place?

Because of 3rd party custom work for OE.

When I looked at this code yesterday it was almost okay. Not following the today's change in the approach.

dew326: That's true we have both. One is for block aligning and the second one is for text alignment in the block. 
And for the custom styles, AlloyEditor sets the text-align.

barw4: @dew326 so where exactly is the problem?

1. When we create content and we add alignment for the custom block it sets - `style="text-align`, even without this PR.
2. When we leave `data-ezalign` the alignment won't work anymore when editing. 
3. When we set  `style="text-align` it behaves exactly as when creating such content. Why is this change invalid though as it just mirrors behavior when creating?

dew326: This change is valid. You should have `text-align` in custom styles and only there as AlloyEditor sets it itself and requires it to work properly.

andrerom: @alongosz It's good that we end up with the right solution here, but FYI edit format is primarily for OE, then secondary to allow imports in html5 as a bonus. So if we have a bug with OE <> format which means we need to change it, then we change it. Same as DB schema, we change *if* we _have_ to.

alongosz: > This change is valid. You should have `text-align` in custom styles and only there as AlloyEditor sets it itself and requires it to work properly.

So now the issue is that Custom Tags stopped working (we've just tested it with @dew326), because the change applied here actually covers both Custom Tags and Custom Styles, since till now they were semantically almost the same for Docbook. `data-ezalign` needs to be kept on Custom Tags, while the current approach is acceptable for custom styles.

> @alongosz It's good that we end up with the right solution here  but FYI edit format is primarily for OE, then secondary to allow imports in html5 as a bonus. So if we have a bug with OE <> format which means we need to change it, then we change it. Same as DB schema, we change _if_ we _have_ to.

Well, ok :+1:  Just don't say I didn't warn you :upside_down_face:  Hence I'll accept the solution as long as it doesn't break Custom Tags.

alongosz (changes requested):
To sum up https://github.com/ezsystems/ezplatform-richtext/pull/166#discussion_r498764268:
Custom Tag alignment with the current changes stopped working.
Custom Style (eztemplate of `"style"` `type`) can have text-align instead of `data-ezalign`.
Custom Tag still needs to have `data-ezalign`.

barw4:
@alongosz resolved in https://github.com/ezsystems/ezplatform-richtext/pull/166/commits/9e86ab387917f2138141ced0a1f0e300ccb0587e

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
andrerom: We will aim to not break custom tags, I didn't mean we should break that. anyway case closed :)

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
```diff
@@ -688,7 +688,17 @@
         <xsl:value-of select="@ezxhtml:class"/>
       </xsl:attribute>
     </xsl:if>
-    <xsl:if test="@ezxhtml:align">
+    <xsl:if test="@xml:id">
+      <xsl:attribute name="id">
+        <xsl:value-of select="@xml:id"/>
+      </xsl:attribute>
+    </xsl:if>
+    <xsl:if test="@ezxhtml:align and not(contains( @ezxhtml:class, 'ez-custom-tag' ))">
```
alongosz, line 696: Hmm, this is unexpected choice of condition for styles. `<eztemplate type="style">` is a custom style and this is what I expected you to rely on when I wrote 

> eztemplate of `"style"` `type`

Should've been more clear I guess, I apologize.

alongosz (changes requested):

src/lib/eZ/RichText/Resources/stylesheets/docbook/xhtml5/edit/core.xsl
barw4: @alongosz hopefully it's gonna be ok now: https://github.com/ezsystems/ezplatform-richtext/pull/166/commits/107f7371670c5f3cf0f21faa8de3aaa533971911

alongosz (approved):

katarzynazawada:
@barw4, storing align and anchor works fine but user cannot link to saved anchor
Steps:

1. Create or edit a content with a rich text field. 
2. In the rich text field, insert a new custom style element.
3. Add anchor on the custom style element.
4. In another paragraph add link to saved anchor.
5. Publish content.

**Result**: Clicking on the link does not lead to the place of adding the anchor.

I am not sure, is it in scope of this PR and customer request or it should be reported separately?

barw4:
@katarzynazawada 

In order for anchor to work properly one must include it in the custom style template like this:
`<div><span id="{{id}}">{{content}}</span></div>`.

katarzynazawada (approved):

lserwatka:
You can merge it up.

barw4:
Merged into 2.1: https://github.com/ezsystems/ezplatform-richtext/commits/2.1
Merged into master: https://github.com/ezsystems/ezplatform-richtext/commit/4e1c61f638c96c0ca6cd01d7eb0755f51f5d7a11

PR #167, opened 2020-10-06
EZP-31853: Fixed inline tags crashing script on creation / edit

Steveb-p:
| Question           | Answer
| **JIRA issue**     | [EZP-31853](https://jira.ez.no/browse/EZP-31853)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 2.0
| **BC breaks**      | no
| **Tests pass**     | yes/no
| **Doc needed**     | no

Inline custom tags expect to work with selection. However, in cases where there were no selection or initially that custom tag was not an inline tag, they would crash the script and cause the editor to become unresponsive.
By adding a check if any node was found this is circumvented.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.


dew326:
It seems like it just silences the problem and not solve it in any way.
If this cannot be done in the backend, we should try validate the configuration on the frontend and do not display buttons in wrong toolbars.

lserwatka:
Well, we can still make sure system is not failing. Thus I would accept this fix until proper solution is in place. We can't have better solution by 3.3.

Steveb-p:
@lserwatka @alongosz Added PoC of a `CompilerPass` that does the required checks. A bit messy from my point of view, but gets the job done.

Is this approach ok? I think adding another parameter to parameter bag would ultimately result in a similar code while increasing the memory footprint ever so slightly for no real gain (since that parameter would imo not be used elsewhere).

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -177,4 +183,72 @@ private function validateCustomTemplatesConfig(
...
+                            sprintf(
+                                'Toolbar "%s" configured in "%s" cannot contain Custom Tag "%s". Inline Custom Tags are not allowed in Toolbars other than "%s".',
+                                $toolbarName,
+                                $siteAccess,
+                                $buttonName,
+                                'text'
+                            )
+                        );
+                    }
+                }
```
alongosz, line 219: CS/Readability: Could you extract some portion of this block to a separate method to reduce nesting level?

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -177,4 +183,72 @@ private function validateCustomTemplatesConfig(
...
+            }
+        }
+    }
+
+    /**
+     * @param array $availableSiteAccesses
+     * @param ContainerBuilder $container
```
alongosz, line 226: CS: either FQCN or actually you can skip strictly defined param & return types if there's no extra information for that parameter (so in this instance @return can stay since it provides extra description).

Side: PhpStorm automatic code analysis not reporting issue is a good indication whether this something a correct choice. Suggested PhpStorm inspection profile for that you can find in ezsystems/ezplatform-code-style#1

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -177,4 +183,72 @@ private function validateCustomTemplatesConfig(
...
+                if ($toolbarName === 'text') {
+                    continue;
+                }
+
+                $buttons = $toolbarContent['buttons'];
+                foreach ($buttons as $buttonName => $buttonConfig) {
+                    if (in_array($buttonName, $customTags, true)) {
+                        throw new InvalidConfigurationException(
+                            sprintf(
+                                'Toolbar "%s" configured in "%s" cannot contain Custom Tag "%s". Inline Custom Tags are not allowed in Toolbars other than "%s".',
```
alongosz, line 211:
```suggestion
                                'Toolbar "%s" configured in "%s" scope cannot contain Custom Tag "%s". Inline Custom Tags are not allowed in Toolbars other than "%s".',
```
proofreading review ping @ezsystems/documentation-team

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -177,4 +183,72 @@ private function validateCustomTemplatesConfig(
...
+    }
+
+    /**
+     * @return string[]
+     */
+    private function getInlineCustomTags(array $customTagsConfig): array
+    {
+        $customTags = array_filter($customTagsConfig, static function (array $customTag): bool {
+            return $customTag['is_inline'] ?? false;
+        });
```
alongosz, line 249: CS/Readability
```suggestion
        $customTags = array_filter(
            $customTagsConfig,
            static function (array $customTag): bool {
                return $customTag['is_inline'] ?? false;
            }
        );
```

alongosz (changes requested):
Looks ok. Can we have some test coverage (unit or Symfony DI extension, there should be some coverage here already in this package).

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
DominikaK: if `scope`, then: `in the "%s" scope`.

Steveb-p: Applied

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
Steveb-p: Extracted the looping over toolbars to a separate method `checkForInlineTagsInToolbar`. I'm not sure if we want to further extract the button check part.

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
Steveb-p: Missed this one, thanks.

alongosz (approved):

dew326 (approved):

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
```diff
@@ -177,4 +182,80 @@ private function validateCustomTemplatesConfig(
...
+    }
+
+    private function checkForInlineTagsInToolbar(
+        string $toolbarName,
+        array $toolbarContent,
+        array $customTags,
+        string $siteAccess
+    ): void {
+        // "text" toolbar is the only one that can contain inline tags
+        if ($toolbarName === 'text') {
```
adamwojs, line 243: Could you please define const. representing toolbar identifiers that can contain inline custom tags?

adamwojs (approved):

src/bundle/DependencyInjection/EzPlatformRichTextExtension.php
Steveb-p: Done.

tomaszszopinski (approved):
QA approved on eZPlatform-ee 3.1.2 with path.

lserwatka:
You can merge it up.

Steveb-p:
@lserwatka merged.

PR #170, opened 2020-12-02
EZP-32194: Disable embed-inline for table/table-row toolbars

SerheyDolgushev:
| Question           | Answer
| **JIRA issue**     | [EZP-32194](https://jira.ez.no/browse/EZP-32194)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 2.2
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

There is no reason to have an "embed-inline" button in the "table" and "table-row" toolbars.

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/config/prepend/ezpublish.yaml
+++ b/src/bundle/Resources/config/prepend/ezpublish.yaml
@@ -298,13 +298,13 @@ system:
                     tr:
                         buttons:
                             ezmoveup:
-                                priority: 60
+                                priority: 70
                             ezmovedown:
-                                priority: 50
+                                priority: 60
                             ezattributesedit:
-                                priority: 40
+                                priority: 50
                             tableHeading:
-                                priority: 30
+                                priority: 40
                             ezanchor:
                                 priority: 20
                             eztableremove:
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/config/prepend/ezpublish.yaml
```diff
@@ -298,14 +298,12 @@ system:
                     tr:
                         buttons:
                             ezmoveup:
-                                priority: 70
-                            ezmovedown:
                                 priority: 60
-                            ezattributesedit:
+                            ezmovedown:
                                 priority: 50
```
dew326, line 303: I don't think we should change the priority of existing buttons, this is kind of a BC break.

dew326 (changes requested):

src/bundle/Resources/config/prepend/ezpublish.yaml
SerheyDolgushev: @dew326 fixed in https://github.com/ezsystems/ezplatform-richtext/pull/170/commits/135a88a81a3826621715fd188fd8f5771a79ef84

dew326 (approved):

GrabowskiM (approved):

lucasOsti (approved):

bogusez (approved):

adamwojs:
Thank you for contribution @SerheyDolgushev!

PR #173, opened 2021-02-09
EZP-32339: As an Editor, I want to edit richtext using CKEditor5

dew326:
| Question           | Answer
| **JIRA issue**     | [EZP-32339](https://jira.ez.no/browse/EZP-32339)
| **Bug/Improvement**| no
| **New feature**    | yes
| **Target version** | feature branch
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [x] Ask for Code Review.


lucasOsti (approved):

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
```diff
@@ -0,0 +1,111 @@
+(function(global, doc, eZ, CKEditor) {
+    class BaseRichText {
+        constructor(config) {
+            this.ezNamespace = 'http://ez.no/namespaces/ezpublish5/xhtml5/edit';
```
GrabowskiM, line 4: ibexa?

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
```diff
@@ -0,0 +1,111 @@
...
+                let i;
+                let newElement;
+
+                if (skipElement) {
+                    newElement = parent;
+                } else {
+                    if (element.nodeType === Node.ELEMENT_NODE) {
+                        newElement = parent.ownerDocument.createElement(element.localName);
+
+                        for (i = 0; i !== element.attributes.length; i++) {
```
GrabowskiM, line 31: why not `for (let i = 0; i !== element.attributes.length; i++) {`? Why declaring this variable at the beginnign?

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
```diff
@@ -0,0 +1,111 @@
...
+                            importChildNodes(newElement, element.attributes[i], false);
```
GrabowskiM, line 32: why not `element.attributes.forEach`?

GrabowskiM (changes requested):

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
dew326: This is what our backend expects and this is not that easy to change. We would have to also change on the backend and probably prepare some migration for previous versions. Definitely out of the scope of this PR.

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
dew326: I think someone suggested this because this `i` is used in two loops.

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
dew326: Because `attributes` is not an array, it array-like object and doesn't have the forEach method.

GrabowskiM (approved):

PR #178, opened 2021-03-01
EZP-32366: As an Editor, I want to use basic features of CKEditor5

dew326:
| Question           | Answer
| **JIRA issue**     | [EZP-32366](https://jira.ez.no/browse/EZP-32366)
| **Bug/Improvement**| no
| **New feature**    | yes
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
+++ b/src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
@@ -1,9 +1,18 @@
 import CharacterCounter from '../plugins/character-counter';
 import InlineEditor from '../../../../../../../../ezplatform-admin-ui-assets/Resources/public/vendors/@ckeditor/ckeditor5-editor-inline/src/inlineeditor';
 import Essentials from '../../../../../../../../ezplatform-admin-ui-assets/Resources/public/vendors/@ckeditor/ckeditor5-essentials/src/essentials';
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
```diff
@@ -99,8 +108,53 @@ import Italic from '../../../../../../../../ezplatform-admin-ui-assets/Resources
...
+                    'italic',
+                    'underline',
+                    'subscript',
+                    'superscript',
+                    'strikethrough',
+                    'blockQuote',
+                ],
+                heading: {
+                    options: [
+                        { model: 'paragraph', title: 'Paragraph', class: 'ck-heading_paragraph' },
```
GrabowskiM, line 146: no translations?

lucasOsti (approved):

src/bundle/Resources/public/js/CKEditor/core/base-ckeditor.js
dew326: These config options will not be here for sure currently don't know where the translation will come from so I would keep it as it is for the time being.

GrabowskiM (approved):

PR #182, opened 2021-04-27
EZP-32400: Fixed displaying anchors for tables

konradoboza:
| Question           | Answer
| **JIRA issue**     | [EZP-32400](https://jira.ez.no/browse/EZP-32400)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `2.2`, `2.3`, `master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Followup for https://github.com/ezsystems/ezplatform-admin-ui/pull/1745 targetting v3. Here, there is also a fix for the wrong anchor icon rendering.

**TODO**:
- [ ] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Resources/public/js/OnlineEditor/core/base-richtext.js
+++ b/src/bundle/Resources/public/js/OnlineEditor/core/base-richtext.js
@@ -156,7 +156,7 @@
             const elementPreviousSibling = element.previousSibling;
             const isTableWithAnchor =
                 element.tagName.toLowerCase() === TABLE_TAG_NAME &&
-                elementPreviousSibling.tagName.toLowerCase() === SVG_TAG_NAME;
+                elementPreviousSibling?.tagName.toLowerCase() === SVG_TAG_NAME;
 
             if (isTableWithAnchor) {
                 elementPreviousSibling.remove();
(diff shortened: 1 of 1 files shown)


dew326 (approved):

adamwojs (approved):

src/bundle/Resources/public/js/OnlineEditor/core/base-richtext.js
```diff
@@ -151,8 +153,14 @@
 
         clearAnchor(element) {
             const icon = element.querySelector('.ez-icon--anchor');
-
-            if (icon) {
+            const elementPreviousSibling = element.previousSibling;
```
Steveb-p, line 156: `element.previousSibling` can be `null`, which would cause the script to crash due to method call on an undefined.

Steveb-p (changes requested):

src/bundle/Resources/public/js/OnlineEditor/core/base-richtext.js
konradoboza: The same comment as in https://github.com/ezsystems/ezplatform-admin-ui/pull/1745#discussion_r621045591.

Steveb-p (approved):

tischsoic (approved):

bogusez (approved):

PR #194, opened 2021-07-05
IBX-707: Improved RichTextConverterExtension libxml-related error handling

konradoboza:
| Question           | Answer
| **JIRA issue**     | [IBX-707](https://issues.ibexa.co/browse/IBX-707)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | `2.3`, `master`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

Given `saveHTML` method might return false in case of external errors/libxml flaws, we should introduce more bulletproof checks for returned values. Error logging doesn't seem to be needed, as the converter will throw an exception in case of any libxml errors anyway, ref: https://github.com/ezsystems/ezplatform-richtext/blob/68e21b4be1477c09bb1a5cf0e2a9fae648a42b69/src/lib/eZ/RichText/Converter/Xslt.php#L140

**TODO**:
- [ ] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/bundle/Templating/Twig/Extension/RichTextConverterExtension.php
+++ b/src/bundle/Templating/Twig/Extension/RichTextConverterExtension.php
@@ -59,7 +59,7 @@ class RichTextConverterExtension extends AbstractExtension
      */
     public function richTextToHtml5(DOMDocument $xmlData): string
     {
-        return $this->richTextOutputConverter->convert($xmlData)->saveHTML();
+        return $this->richTextOutputConverter->convert($xmlData)->saveHTML() ?: '';
     }
 
     /**
@@ -71,6 +71,6 @@ class RichTextConverterExtension extends AbstractExtension
      */
     public function richTextToHtml5Edit(DOMDocument $xmlData): string
     {
-        return $this->richTextEditConverter->convert($xmlData)->saveHTML();
+        return $this->richTextEditConverter->convert($xmlData)->saveHTML() ?: '';
     }
 }


adamwojs (approved):

mateuszdebinski (approved):

Steveb-p (approved):
Does it make sense to have tests for these?

There's also the possibility of changing the `saveHTML` method to always return `string` on it's own, but I guess there's value in it returining `false` in some cases? (despite the possible exceptions).

mateuszdebinski:
we have a lot of places that call richtext convert and save HTML, shouldn't we introduce this everywhere (where it needs to be, of course)? for example in: `ezsystems/ezplatform-page-fieldtype/src/lib/GraphQL/Resolver/RichtextBlockAttributeTypeResolver.php`?

adamwojs:
> There's also the possibility of changing the saveHTML method to always return string on it's own, but I guess there's value in it returining false in some cases? (despite the possible exceptions).

Yes, but this will require contribution to https://github.com/php/php-src. It's https://www.php.net/manual/en/domdocument.savehtml.php :wink:

konradoboza:
@mateuszdebinski I think this could be indeed introduced in other repositories but in methods that directly use `string` as a type-hint. WDYT @adamwojs @alongosz?

Steveb-p:
> > There's also the possibility of changing the saveHTML method to always return string on it's own, but I guess there's value in it returining false in some cases? (despite the possible exceptions).
> 
> Yes, but this will require contribution to https://github.com/php/php-src. It's https://www.php.net/manual/en/domdocument.savehtml.php

I assumed we wrap it with our own class (since I've noticed we're using our own interface).

alongosz (changes requested):
1. Missing test coverage
2. Missing logging of an error. It might be more clear how to log this if we have 1.

Steveb-p:
> 2\. Missing logging of an error. It might be more clear how to log this if we have 1.

@konradoboza it can be done easily by using `LoggerAwareInterface` and `LoggerAwareTrait` btw.

see https://symfony.com/doc/current/logging.html#using-a-logger-inside-a-service

patrickallaert:
> There's also the possibility of changing the `saveHTML` method to always return `string` on it's own, but I guess there's value in it returining `false` in some cases? (despite the possible exceptions).

Not going to happen any time soon (the earliest would be PHP 9.0) because of serious BC break.
A `false` being returned really denotes something wrong happened (unless you distribution provides you libxml2 2.9.12 without a patch for https://gitlab.gnome.org/GNOME/libxml2/-/issues/266 when you try converting an empty DOM to HTML).

patrickallaert:
> 1. Missing test coverage

This might be very tricky to get as it requires a buggy version of libxml2 that would enables PHP's `DOMDocument->saveHTML()` to return `false`.

alongosz:
> > 1. Missing test coverage
> 
> This might be very tricky to get as it requires a buggy version of libxml2 that would enables PHP's `DOMDocument->saveHTML()` to return `false`.

@patrickallaert I was rather thinking of loading erroneous document that would return that. Since the behavior is documented I assumed there are specific cases when it would happen apart from buggy libxml version.

alongosz (approved):
After internal sync with @konradoboza looks like we've expired possibilities of making proper logging & coverage and looks like the erroneous state is related indeed only to specific libxml version(s)

Hence, +1

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3.6 & 3.3.6-dev with patch & diff.

PR #198, opened 2021-07-14
IBX-585: Implemented support for inline custom tags

dew326:
| Question           | Answer
| **JIRA issue**     | [IBX-585](https://jira.ez.no/browse/IBX-585)
| **New feature**    | yes
| **Target version** | master
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [x] Ask for Code Review.


src/bundle/Resources/public/js/CKEditor/custom-tags/custom-tags.js
```diff
@@ -1,11 +1,29 @@
...
+                    this.componentName = name;
+                    this.config = config;
+
+                    this.formView.setChildren({
+                        attributes: this.config.attributes,
+                    });
+                }
+            };
+        });
+        const customTagsUi = blockCustomTags.map(([name, config]) => {
```
tischsoic, line 26: Maybe similarily to `inlineCustomTagsUi`:
```suggestion
        const blockCustomTagsUi = blockCustomTags.map(([name, config]) => {
```

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-editing.js
```diff
@@ -0,0 +1,146 @@
...
+                    return;
+                }
+
+                const configElement = viewElement.getChild(1);
+                const configValuesIterator = configElement.getChildren();
+                const customTagName = viewElement.getAttribute('data-ezname');
+                const values = {};
+
+                for (const configValue of configValuesIterator) {
+                    const value = (configValue.getChild(0) && configValue.getChild(0).data) || null;
```
tischsoic, line 90: What about:
```suggestion
                    const value = configValue.getChild(0)?.data ?? null;
```
It works a little bit differently because e.g. it would return `0` instead of `null` in case `data` is equal to `0`. But do we want to return null in case `data` is `0`? Or maybe `data` is always a string etc? Could you explain?

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-ui.js
```diff
@@ -0,0 +1,150 @@
...
+            this.hideForm();
+        });
+
+        return formView;
+    }
+
+    showForm() {
+        const modelElement = this.editor.model.document.selection.getSelectedElement();
+        const values = modelElement.getAttribute('values');
+        console.log('show', values);
```
tischsoic, line 81: `console.log`?

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-ui.js
```diff
@@ -0,0 +1,150 @@
...
+        return { target: view.domConverter.viewRangeToDom(range) };
+    }
+
+    addInlineCustomTag() {
+        const values = Object.entries(this.config.attributes).reduce((values, [attributeName, config]) => {
+            values[attributeName] = config.defaultValue;
+
+            return values;
+        }, {});
+        console.log(values);
```
tischsoic, line 120: `console.log`?

tischsoic (changes requested):
Requesting changes because of 2 `console.log`s which, I guess, should be deleted. The rest for me is OK. 👍  🙂

src/bundle/Resources/public/js/CKEditor/custom-tags/custom-tags.js
```diff
@@ -1,11 +1,29 @@
...
+import IbexaInlineCustomTagsUI from './inline-custom-tag/inline-custom-tag-ui';
+import IbexaCustomTagsEditing from './block-custom-tag/custom-tag-editing';
+import IbexaInlineCustomTagsEditing from './inline-custom-tag/inline-custom-tag-editing';
 
 class IbexaCustomTags extends Plugin {
     static get requires() {
-        const customTagsUi = Object.entries(window.eZ.richText.customTags).map(([name, config]) => {
+        const blockCustomTags = Object.entries(window.eZ.richText.customTags).filter(([name, config]) => !config.isInline);
+        const inlineCustomTags = Object.entries(window.eZ.richText.customTags).filter(([name, config]) => config.isInline);
+        const inlineCustomTagsUi = inlineCustomTags.map(([name, config]) => {
```
GrabowskiM, line 12:
```suggestion
        const inlineCustomTagsUI = inlineCustomTags.map(([name, config]) => {
```

src/bundle/Resources/public/js/CKEditor/custom-tags/custom-tags.js
GrabowskiM:
```suggestion
        const customTagsUI = blockCustomTags.map(([name, config]) => {
```

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-editing.js
```diff
@@ -0,0 +1,146 @@
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import { toWidget } from '@ckeditor/ckeditor5-widget/src/utils';
```
GrabowskiM, line 2: what would you say on moving this line below? I personally prefer having import utils below imports of classes

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-editing.js
```diff
@@ -0,0 +1,146 @@
...
+            view: (modelElement, { writer: downcastWriter }) => {
+                const container = downcastWriter.createContainerElement('span', {
+                    'data-ezelement': 'eztemplateinline',
+                    'data-ezname': modelElement.getAttribute('customTagName'),
+                });
+                const config = downcastWriter.createUIElement('span', { 'data-ezelement': 'ezconfig' }, function(domDocument) {
+                    const domElement = this.toDomElement(domDocument);
+
+                    domElement.innerHTML = Object.entries(modelElement.getAttribute('values')).reduce((total, [attribute, value]) => {
+                        const attributeValue = value !== null ? value : '';
```
GrabowskiM, line 56: You can use `value ?? ''` here.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
```
The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.
```

tischsoic (approved):

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-editing.js
```diff
@@ -0,0 +1,146 @@
...
+                if (viewElement.getAttribute('data-eztype') === 'style') {
+                    return;
+                }
+
+                const configElement = viewElement.getChild(1);
+                const configValuesIterator = configElement.getChildren();
+                const customTagName = viewElement.getAttribute('data-ezname');
+                const values = {};
+
+                for (const configValue of configValuesIterator) {
```
GrabowskiM, line 89: Is it possible to use reduce here? `[...configValuesIterator].reduce()` I think

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-ui.js
```diff
@@ -0,0 +1,150 @@
...
+    constructor(props) {
+        super(props);
+
+        this.balloon = this.editor.plugins.get('ContextualBalloon');
+        this.formView = this.createFormView();
+
+        this.showForm = this.showForm.bind(this);
+        this.addInlineCustomTag = this.addInlineCustomTag.bind(this);
+
+        this.isNew = false;
```
GrabowskiM, line 18: Here we go again... ;P

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-ui.js
```diff
@@ -0,0 +1,150 @@
...
+        });
+    }
+
+    createFormView() {
+        const formView = new IbexaCustomTagFormView({ locale: this.editor.locale });
+
+        this.listenTo(formView, 'save-custom-tag', () => {
+            const modelElement = this.editor.model.document.selection.getSelectedElement();
+            const values = modelElement.getAttribute('values');
+            const newValues = Object.assign({}, values);
```
GrabowskiM, line 50:
```suggestion
            const newValues = {...values};
```

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-ui.js
```diff
@@ -0,0 +1,150 @@
...
+    getBalloonPositionData() {
+        const view = this.editor.editing.view;
+        const viewDocument = view.document;
+        const range = viewDocument.selection.getFirstRange();
+
+        return { target: view.domConverter.viewRangeToDom(range) };
+    }
+
+    addInlineCustomTag() {
+        const values = Object.entries(this.config.attributes).reduce((values, [attributeName, config]) => {
```
GrabowskiM, line 115: I'm not fan of using two different `values`, but in that case I guess it's ok

GrabowskiM (changes requested):

src/bundle/Resources/public/js/CKEditor/custom-tags/inline-custom-tag/inline-custom-tag-editing.js
dew326: I don't know if there is a point of doing that (conversion from iterator to array)

lucasOsti (approved):

PR #225, opened 2022-04-27
IBX-2814: Returned null instead of empty array when there are no child nodes

ViniTou:
| Question           | Answer
| **JIRA issue**     | https://issues.ibexa.co/browse/IBX-2814
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | latest stable for bug fixes, master for new features
| **BC breaks**      | **yes**
| **Tests pass**     | yes
| **Doc needed**     | yes (this should probably be mentioned somewhere)

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Converter/Render.php
+++ b/src/lib/eZ/RichText/Converter/Render.php
@@ -55,11 +55,15 @@ abstract class Render
      *
      * @param \DOMNode $configHash
      *
-     * @return array
+     * @return array|null
      */
     protected function extractHash(DOMNode $configHash)
     {
         $hash = [];
+
+        if ($configHash->childNodes->count() === 0) {
+            return null;
+        }
 
         foreach ($configHash->childNodes as $node) {
             /** @var \DOMText|\DOMElement $node */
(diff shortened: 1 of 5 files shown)


lserwatka (approved):

dew326:
Shouldn't we change checks in our custom tags twig templates? I know this is not a case in this version because AlloyEditor works differently than CKEditor but I think we should have twigs prepared for null.

adamwojs (approved):

konradoboza (approved):
Hence this PR is targetting `ezplatform-richtext` I assume version 3.3 is also affected - worth mentioning in the ticket.

ViniTou:
@dew326 
It works fine with current state of templates, as it simply adds empty html attribute. 

@konradoboza @micszo 
This issue is not reproducible on 3.3 as, Darek mentioned, AlloyEditor skips empty keys. But since it is technically possible to replace AE with CKEditor, and overall that solution helps us avoiding strange situations generally, I fixed it at 3.3.
So at 3.3, all you need to check if that it didnt broke anything.

micszo (approved):
Retested on Ibexa DXP v3.3 & v4.

PR #233, opened 2023-06-14
IBX-5961: Fixed validation error due to not accepted TextAlign value

vidarl:
| Question           | Answer
| **JIRA issue**     | [IBX-5961](https://jira.ez.no/browse/IBX-5961)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | Ibexa DXP 3.3 and 4.5
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

When cut&pasting from word, you might get TextAlign values that doesn't match our [schema](https://github.com/ezsystems/ezplatform-richtext/blob/v2.3.19/src/lib/eZ/RichText/Resources/schemas/docbook/ezpublish.rng#L777-L780).

With this patch, TextAlign will default to"left" if the value specified is not acceptable

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.

### DIFF
--- a/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
+++ b/src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
@@ -99,7 +99,7 @@
       </xsl:if>
       <xsl:if test="contains( @style, 'text-align:' )">
         <xsl:variable name="textAlign">
-          <xsl:call-template name="extractStyleValue">
+          <xsl:call-template name="extractTextAlignValue">
             <xsl:with-param name="style" select="@style"/>
             <xsl:with-param name="property" select="'text-align'"/>
           </xsl:call-template>
@@ -314,7 +314,7 @@
       </xsl:if>
       <xsl:if test="contains( @style, 'text-align:' )">
         <xsl:variable name="textAlign">
-          <xsl:call-template name="extractStyleValue">
+          <xsl:call-template name="extractTextAlignValue">
             <xsl:with-param name="style" select="@style"/>
             <xsl:with-param name="property" select="'text-align'"/>
           </xsl:call-template>
@@ -728,7 +728,7 @@
       </xsl:if>
       <xsl:if test="contains( @style, 'text-align:' )">
         <xsl:variable name="textAlign">
-          <xsl:call-template name="extractStyleValue">
+          <xsl:call-template name="extractTextAlignValue">
             <xsl:with-param name="style" select="@style"/>
             <xsl:with-param name="property" select="'text-align'"/>
           </xsl:call-template>
@@ -785,6 +785,24 @@
     <xsl:param name="style"/>
     <xsl:param name="property"/>
     <xsl:value-of select="translate( substring-before( substring-after( concat( substring-after( $style, $property ), ';' ), ':' ), ';' ), ' ', '' )"/>
+  </xsl:template>
+
+  <xsl:template name="extractTextAlignValue">
+    <xsl:param name="style"/>
+    <xsl:param name="property"/>
+    <xsl:variable name="alignValue">
+      <xsl:call-template name="extractStyleValue">
+        <xsl:with-param name="style" select="$style"/>
+        <xsl:with-param name="property" select="$property"/>
+      </xsl:call-template>
+    </xsl:variable>
+
+    <xsl:choose>
+      <xsl:when test="$alignValue = 'left' or $alignValue = 'center' or $alignValue = 'right' or $alignValue = 'justify'">
+        <xsl:value-of select="$alignValue"/>
+      </xsl:when>
+      <xsl:otherwise>left</xsl:otherwise>
+    </xsl:choose>
   </xsl:template>
 
   <xsl:template name="normalizeWidth">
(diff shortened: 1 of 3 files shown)


src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -787,6 +787,24 @@
...
+        <xsl:with-param name="style" select="$style"/>
+        <xsl:with-param name="property" select="$property"/>
+      </xsl:call-template>
+    </xsl:variable>
+
+    <xsl:choose>
+      <xsl:when test="$alignValue = 'left' or $alignValue = 'center' or $alignValue = 'right' or $alignValue = 'justify'">
+        <xsl:value-of select="$alignValue"/>
+      </xsl:when>
+      <xsl:otherwise>left</xsl:otherwise>
```
vidarl, line 804: So, I here a default value of `left` is used when the value provided is not supported.

This means that we silently changes whatever user provided. However, the validation errors provided by libxml is anyway not very informative so I think this is okay

vidarl:
@ezsystems/php-dev-team : please review

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
```diff
@@ -787,6 +787,24 @@
...
+    <xsl:param name="property"/>
+    <xsl:variable name="alignValue">
+      <xsl:call-template name="extractStyleValue">
+        <xsl:with-param name="style" select="$style"/>
+        <xsl:with-param name="property" select="$property"/>
+      </xsl:call-template>
+    </xsl:variable>
+
+    <xsl:choose>
+      <xsl:when test="$alignValue = 'left' or $alignValue = 'center' or $alignValue = 'right' or $alignValue = 'justify'">
```
konradoboza, line 801: Isn't there any way to iterate over alignment directions coming from `ezpublish.rng`? This will make it a bit more bulletproof whenever schema changes.

konradoboza (changes requested):
Please provide some test coverage showing that `text-align: left` is indeed set by default.

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
vidarl: Maybe...
It is possible to load other documents from the xsl stylesheet.
However, to check such things against the schema inside the stylesheet  will add a completely new level of complexity so I wouldn't recommend it.

Anyway, upon changes to the schema, new tests should be written and those should reveal this hard coded limit

vidarl:
> Please provide some test coverage showing that `text-align: left` is indeed set by default.

Ahh.. two file are missing from the commit.. comming right up...

tests/lib/eZ/RichText/Converter/Xslt/_fixtures/xhtml5/edit/lossy/009-ezxhtmlTextalignInvalid.docbook.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<section xmlns="http://docbook.org/ns/docbook" xmlns:ezcustom="http://ez.no/xmlns/ezpublish/docbook/custom" xmlns:ezxhtml="http://ez.no/xmlns/ezpublish/docbook/xhtml" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0-variant ezpublish-1.0">
+    <para ezxhtml:textalign="left">Some text</para>
+</section>
```
konradoboza:
```suggestion
</section>

```

konradoboza (approved):
LGTM concerning challenges coming from converting inconsistent Word-related formatting. Probably @alongosz might have some further remarks.

src/lib/eZ/RichText/Resources/stylesheets/xhtml5/edit/docbook.xsl
konradoboza: That should be fine, as there are no other values for `text-align` attribute AFAICS and this shouldn't change anytime soon.

alongosz: > Isn't there any way to iterate over alignment directions coming from `ezpublish.rng`? This will make it a bit more bulletproof whenever schema changes.



> Maybe... It is possible to load other documents from the xsl stylesheet. However, to check such things against the schema inside the stylesheet will add a completely new level of complexity so I wouldn't recommend it.
> 
> Anyway, upon changes to the schema, new tests should be written and those should reveal this hard coded limit

I wouldn't do that. XSL is a document describing transformation from one format to another. `ezpublish.rng` is a RelaxNG schema for DocBook format. Moreover OOTB `xsl:import` should (can?) import only XSL.

It's fine as is, given this is still XSLT 1.0.

alongosz (approved):
I still believe that input sanitization is a responsibility of an input provider, not handler, which is Online Editor in this case. We're opening ourselves here to wide range of bug fixes which might in the end pollute already complicated XSLT.

But given this solves the issue in a clean way (at least for XSL standards): +1

micszo (approved):
QA Approved on Ibexa Commerce 3.3.34-dev.

PR #237, opened 2023-07-10
IBX-5048: Fixed handling nested lists with line feeds

vidarl:
| Question           | Answer
| **JIRA issue**     | [IBX-5048](https://issues.ibexa.co/browse/IBX-5048)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 2.3
| **BC breaks**      | no
| **Tests pass**     | yes/no
| **Doc needed**     | no

`<itemizedList>` and `<orderedList>` cannot be below `<literallayout>` in richtext. It was quite hard to deal with that in XSLT, so created a php converter to deal with it after XSLT has completed

**TODO**:
- [x] Implement tests.
- [x] Ask for Code Review.


phpstan-baseline.neon
```diff
@@ -560,6 +560,11 @@ parameters:
 			count: 1
 			path: src/lib/Form/Type/RichTextFieldType.php
 
+		-
+			message: "#^Variable \\$element in PHPDoc tag @var does not match any variable in the foreach loop\\: \\$childNode$#"
+			count: 1
+			path: src/lib/RichText/Converter/ListLineBreaks.php
```
vidarl: FYI : Added for the same reasons as d3ac5fad35fd219f5b958557e19d7069759df18c

vidarl:
@ezsystems/php-dev-team : Please review

composer.json
```diff
@@ -45,14 +45,16 @@
     "autoload": {
         "psr-4": {
             "EzSystems\\EzPlatformRichTextBundle\\": "src/bundle",
-            "EzSystems\\EzPlatformRichText\\": "src/lib"
+            "EzSystems\\EzPlatformRichText\\": "src/lib",
+            "Ibexa\\FieldTypeRichText\\RichText\\": "src/lib/RichText"
```
ViniTou:
```suggestion
            "Ibexa\\FieldTypeRichText\\RichText\\": "src/lib/RichText",
            "EzSystems\\EzPlatformRichText\\": "src/lib"
```

Adding new ones on top helps with phpstorm suggestions.

src/bundle/Resources/config/fieldtype_services.yaml
```diff
@@ -139,6 +139,12 @@ services:
         tags:
             - {name: ezrichtext.converter.input.xhtml5, priority: 10}
 
+    # Note: should run after xsl transformation
+    ezrichtext.converter.input.xhtml5.listlinebreaks:
```
ViniTou: Can we just use FQCN as service name?
```suggestion
    Ibexa\FieldTypeRichText\RichText\Converter\ListLineBreaks:
```

ViniTou (approved):

src/bundle/Resources/config/fieldtype_services.yaml
Nattfarinn: Even if we need it in a form of `ezrichtext.converter.input.xhtml5.listlinebreaks`, I would just change it anyway and used an alias. In either case, remove the `class:` entry as well

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+        $elements = $xpath->query($xpathExpression) ?: [];
+
+        // elements are list of <literallayout> elements
+        foreach ($elements as $element) {
+            $listItemToMove = [];
+            $listItemNo = 0;
+            /** @var \DOMElement $element */
+            foreach ($element->childNodes as $childNode) {
+                /** @var \DOMNode $childNode */
+                if ($childNode instanceof \DOMElement) {
```
Nattfarinn: Use `use`, you already have `DOMDocument` and `DOMXPath` there.

Nattfarinn (approved):

tests/lib/Richtext/Converter/ListLineBreaksTest.php
```diff
@@ -0,0 +1,96 @@
...
+
+use EzSystems\EzPlatformRichText\eZ\RichText\Converter\Link;
+use Ibexa\FieldTypeRichText\RichText\Converter\ListLineBreaks;
+use PHPUnit\Framework\TestCase;
+use DOMDocument;
+
+/**
+ * Tests the Link converter
+ * Class LinkTest.
+ */
```
alongosz: This comment doesn't carry useful information.
Please replace it by `@covers \<FQCN>\ListLineBreaks`

tests/lib/Richtext/Converter/ListLineBreaksTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $inputDocument->loadXML($input);
+
+        $converter = new ListLineBreaks();
+
+        $outputDocument = $converter->convert($inputDocument);
+
+        $expectedOutputDocument = new DOMDocument();
+        $expectedOutputDocument->loadXML($output);
+
+        $this->assertEquals($expectedOutputDocument, $outputDocument);
```
alongosz:
```suggestion
        self::assertEquals($expectedOutputDocument, $outputDocument);
```

tests/lib/Richtext/Converter/ListLineBreaksTest.php
```diff
@@ -0,0 +1,96 @@
...
+use EzSystems\EzPlatformRichText\eZ\RichText\Converter\Link;
+use Ibexa\FieldTypeRichText\RichText\Converter\ListLineBreaks;
+use PHPUnit\Framework\TestCase;
+use DOMDocument;
+
+/**
+ * Tests the Link converter
+ * Class LinkTest.
+ */
+class ListLineBreaksTest extends TestCase
```
alongosz:
```suggestion
final class ListLineBreaksTest extends TestCase
```

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+                    if ($childNode->tagName == 'orderedlist' || $childNode->tagName == 'itemizedlist') {
+                        $listItemToMove[$listItemNo] = $childNode;
```
alongosz: Wouldn't `$listItemToMove[] = ` be enough?

I'm curious however why do you need to create list of nodes instead of performing code from L56-60 on the fly here?

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+use EzSystems\EzPlatformRichText\eZ\RichText\Converter;
+use DOMDocument;
+use DOMXPath;
+
+/**
+ * Class ListLineBreaks.
+ *
+ * Processes lists below <literallayout> tag.
+ */
+class ListLineBreaks implements Converter
```
alongosz: Naming & PHPDoc & extensibility:
```suggestion
/**
 * @internal
 *
 * Processes lists nested in <literallayout> tags.
 */
final class LiteralLayoutNestedList implements Converter
```
Intentionally skipping LineBreak(s) as this is too much of an implementation detail. It's just a converter handling nested lists in `<literallayout>`.

_Side note: I would suffix the name with Converter just to distinguish it from potential ValueObject, but that's inconsistent with the other Converter and there are probably different opinions about that, so not insisting ;)_

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+
+        // elements are list of <literallayout> elements
+        foreach ($elements as $element) {
+            $listItemToMove = [];
+            $listItemNo = 0;
+            /** @var \DOMElement $element */
+            foreach ($element->childNodes as $childNode) {
+                /** @var \DOMNode $childNode */
+                if ($childNode instanceof \DOMElement) {
+                    if ($childNode->tagName == 'orderedlist' || $childNode->tagName == 'itemizedlist') {
```
alongosz:
```suggestion
                    if ($childNode->tagName === 'orderedlist' || $childNode->tagName === 'itemizedlist') {
```

phpstan-baseline.neon
alongosz: > FYI : Added for the same reasons as [d3ac5fa](https://github.com/ezsystems/ezplatform-richtext/commit/d3ac5fad35fd219f5b958557e19d7069759df18c)

This is not related to d3ac5fad35fd219f5b958557e19d7069759df18c. 

If you change the relevant code to state:
```php
        /** @var \DOMElement $element */
        foreach ($elements as $element) {
```
the issue will disappear. Otherwise PHPStan seems to assume you're trying to type-hint the nested `foreach` right after that statement.

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+        $xpathExpression = '//ns:literallayout [descendant::ns:orderedlist|descendant::ns:itemizedlist]';
+        $ns = $document->documentElement ? $document->documentElement->namespaceURI ?: '' : '';
+        $xpath->registerNamespace('ns', $ns);
+        $elements = $xpath->query($xpathExpression) ?: [];
+
+        // elements are list of <literallayout> elements
+        foreach ($elements as $element) {
+            $listItemToMove = [];
+            $listItemNo = 0;
+            /** @var \DOMElement $element */
```
alongosz: As mentioned earlier, try moving this `@var` type-hint it directly above `foreach` in L38 and see if PHPStan issue.

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+            foreach ($element->childNodes as $childNode) {
+                /** @var \DOMNode $childNode */
```
alongosz: Let's also do for consistency's sake:
```suggestion
            /** @var \DOMNode $childNode */
            foreach ($element->childNodes as $childNode) {                
```
_Not sure if I agree with type-hinting before the variable gets created, but this seems to be consistent expectation based on my experience with both PHPStan and PHPStorm static analysis tools_

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+     * @param \DOMDocument $document
+     *
+     * @return \DOMDocument
+     */
+    public function convert(DOMDocument $document)
+    {
+        $xpath = new DOMXPath($document);
+        $xpathExpression = '//ns:literallayout [descendant::ns:orderedlist|descendant::ns:itemizedlist]';
+        $ns = $document->documentElement ? $document->documentElement->namespaceURI ?: '' : '';
+        $xpath->registerNamespace('ns', $ns);
```
alongosz: We shouldn't register empty namespace.
```php
private const FALLBACK_NAMESPACE = 'http://docbook.org/ns/docbook';
```
```suggestion
        $xpath->registerNamespace(
            'ns',
            null !== $document->documentElement && !empty($document->documentElement->namespaceURI)
                ? $document->documentElement->namespaceURI
                : self::FALLBACK_NAMESPACE
        );
```

src/lib/RichText/Converter/ListLineBreaks.php
```diff
@@ -0,0 +1,67 @@
...
+                    if ($targetNode !== null) {
+                        $targetNode->appendChild($node);
+                    }
+                }
+            }
+        }
+
+        return $document;
+    }
+}
```
alongosz: @vidarl I apologize for a bit chaotic review. I see now that it might be difficult to address it so I summarized my suggestions & more as the diff. Let me know if that works for you or if you have any questions.

```suggestion
use DOMDocument;
use DOMElement;
use DOMNode;
use DOMXPath;
use EzSystems\EzPlatformRichText\eZ\RichText\Converter;

/**
 * @internal
 *
 * Processes lists nested in <literallayout> tags.
 */
final class LiteralLayoutNestedList implements Converter
{
    private const FALLBACK_NAMESPACE = 'http://docbook.org/ns/docbook';
    private const ORDERED_LIST_TAG = 'orderedlist';
    private const ITEMIZED_LIST_TAG = 'itemizedlist';

    /**
     * For all <itemizedList> and <orderedList> nested in the <literallayout>, move the list after the <literallayout>,
     * so that it is not inside it.
     */
    public function convert(DOMDocument $document): DOMDocument
    {
        $xpath = new DOMXPath($document);
        $xpathExpression = '//ns:literallayout [descendant::ns:orderedlist|descendant::ns:itemizedlist]';
        $xpath->registerNamespace(
            'ns',
            null !== $document->documentElement && !empty($document->documentElement->namespaceURI)
                ? $document->documentElement->namespaceURI
                : self::FALLBACK_NAMESPACE
        );
        $elements = $xpath->query($xpathExpression) ?: [];

        // elements are list of <literallayout> elements
        /** @var \DOMElement $element */
        foreach ($elements as $element) {
            /** @var \DOMNode $childNode */
            foreach ($element->childNodes as $childNode) {
                if ($this->isNestedListNode($childNode)) {
                    $targetNode = $childNode->parentNode->parentNode;
                    if ($targetNode !== null) {
                        $targetNode->appendChild($childNode);
                    }
                }
            }
        }

        return $document;
    }

    /**
     * @phpstan-assert-if-true !null $childNode->parentNode
     */
    private function isNestedListNode(DOMNode $childNode): bool
    {
        return $childNode instanceof DOMElement
            && ($childNode->tagName === self::ORDERED_LIST_TAG || $childNode->tagName === self::ITEMIZED_LIST_TAG)
            && $childNode->parentNode !== null;
    }
}
```

alongosz (commented):
@vidarl please see first my last diff comment, it contains all the suggested changes for ListLineBreaks class from other comments to make it more clear.

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Because you cannot change a DOMNodeList while you are looping over it.
[This comment](https://www.php.net/manual/en/class.domnodelist.php#83180) is old, but is still valid it seems

alongosz: > Because you cannot change a DOMNodeList while you are looping over it. [This comment](https://www.php.net/manual/en/class.domnodelist.php#83180) is old, but is still valid it seems

@vidarl if you see the full snippet I proposed at the end of the last review, it does't seem to make a difference if move node while looping or on the fly. So the question is - do we have some missing test coverage or is the comment actually not valid anymore? Could you provide an example input and output DOM for which the code would fail?

src/bundle/Resources/config/fieldtype_services.yaml
vidarl: I used the old-style naming in order to be consistent with existing code. Will change to FQCN and remove class then

vidarl: Fixed in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

tests/lib/Richtext/Converter/ListLineBreaksTest.php
vidarl: Fixed in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

phpstan-baseline.neon
vidarl: ok. Thanx

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

tests/lib/Richtext/Converter/ListLineBreaksTest.php
vidarl: Fixed in d40261975b63eb225f11af48600811c3b16d548f

tests/lib/Richtext/Converter/ListLineBreaksTest.php
vidarl: Fixed in d40261975b63eb225f11af48600811c3b16d548f

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed as suggested in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed as suggested in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed as suggested in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed as suggested in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: Fixed as suggested in 295b2127692c3b2b2aaa26c0c9310cb3c916a2f4

src/lib/RichText/Converter/ListLineBreaks.php
vidarl: While working on this, or maybe some other PR I experienced that messing with a DOMNodeList while looping over it caused problems. From that point I have just been very careful about the mater and created a second loop just to be safe. I tested your code now and it works fine so I think we are good then.

vidarl:
I believe all review comments have been addressed now. Feel free to have a final look

alongosz (approved):

kisztof (approved):

jwibexa (approved):
tested and validated. 
Please take note that the PR requires additional line in composer.json:
```
    "autoload": {
        "psr-4": {
            "App\\": "src/",
+            "Ibexa\\FieldTypeRichText\\RichText\\": "vendor/ezsystems/ezplatform-richtext/src/lib/RichText"
        }
    },
```

PR #244, opened 2024-09-18
IBX-8891: xml link validator in rich text does not allow to use "tel:" links

vidarl:
| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | [IBX-8891](https://jira.ez.no/browse/IBX-8891)
| **Bug/Improvement**| yes
| **New feature**    | no
| **Target version** | 3.3, 4.6, main
| **BC breaks**      | no
| **Tests pass**     | no
| **Doc needed**     | no

This is an attempt to fix [ibexa-sa-2024-005-persistent-xss-in-richtext](https://developers.ibexa.co/security-advisories/ibexa-sa-2024-005-persistent-xss-in-richtext) without touching original docbook schema.
However, there are some challenges, see review comment by author

**TODO**:
- [x] Implement feature / fix a bug.
- [x] Implement tests.
- [x] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [ ] Ask for Code Review.

PR #245, opened 2024-09-19
IBX-8891: Fixed XML link validation to allow using `tel:` in links

vidarl:
| :ticket: Issue | [IBX-8891](https://issues.ibexa.co/browse/IBX-8891) |
|----------------|-----------|

#### Description:

Adds supports for 
 - `tel: ` links
 - Remove check for quotes and < and >, output is anyway url encoded

I synced with Gunnstein and we agree to keep the validation in schematron, but to simplify it, in order to keep the number of checks down as there is a default limit of 9:
```
runtime error: file /home/foobar/git/ezplatform-richtext/src/lib/eZ/RichText/Resources/stylesheets/schematron/iso_dsdl_include.xsl line 1353 element choose xsltApplySequenceConstructor: A potential infinite template recursion was detected.
You can adjust xsltMaxDepth (--maxdepth) in order to raise the maximum number of nested template calls and variables/params (currently set to 3000).
```
 The most important think in this case is to ensure the security aspect, not to ensure that the URLs are syntactical correct (ie, we don't need to check for `ezcontent://`, `ezlocation://`, `ezremote://` or `ezurl://`. We can simply have one rule that check for "starts with ez")

You may have a look at https://github.com/ezsystems/ezplatform-richtext/pull/244 for a different approach


glye (approved):

phpstan-baseline.neon
```diff
@@ -625,6 +625,36 @@ parameters:
 			count: 1
 			path: src/lib/eZ/FieldType/RichText/RichTextStorage.php
 
+		-
+			message: "#^Offset 1 does not exist on array\\{0\\?\\: string, 1\\?\\: ''\\|'ezremote\\://', 2\\?\\: string, 3\\?\\: string\\}\\.$#"
+			count: 1
+			path: src/lib/eZ/FieldType/RichText/RichTextStorage.php
```
alongosz, line 631: Note: allowing this since this is baseline alignment after an update and fixing this requires some changes that would take significant amount of time (impl + QA)

alongosz (approved):

mikadamczyk (approved):

phpstan-baseline.neon
vidarl: yeah, and the same thing was done in 4.6 : https://github.com/ibexa/fieldtype-richtext/commit/560eecf461c7bbba36dd1d1df9fb8bcf338ad34f

konradoboza (approved):

tomaszszopinski (approved):
QA approved on IbexaDXP 3.3 commerce.

